PR #7, opened 2017-05-23
Add details to threads proposal agenda

binji:


binji:
Not sure whether to add [Consider supporting verifiably safe concurrency](https://github.com/WebAssembly/threads/issues/3) to the list.

jfbastien:
> Not sure whether to add [Consider supporting verifiably safe concurrency](https://github.com/WebAssembly/threads/issues/3) to the list.

We could discuss briefly and see if there's interest in pursuing it for v1 of threads, or if we believe someone could champion this later.

binji:
Good idea, I've added it in.

jfbastien:
I think you want to get official guidance from the committee on specific design items for threads. A few ideas:

* Do we like that it matches the SAB design?
* What about blocking main thread?
* Do we want a way to create threads native to wasm?
* Should that way be in v1 or later?
* Do we want `i64` atomics, and `await`?

Some of these questions you're pretty sure of the answer you'll get. That's great! Consensus means we won't revisit unless new information comes to light. If you get a surprising answer then great!

Further, you want to ask about open work items for which you need guidance. You have 2 in there, I think there are more?

binji:
> I think you want to get official guidance from the committee on specific design items for threads.

These sound good, is it worth adding new issues for them, or just add them to the agenda?

> Further, you want to ask about open work items for which you need guidance. You have 2 in there, I think there are more?

I took a quick look over the design issues and couldn't find any. Are there any you're thinking of in particular?

jfbastien:
>> I think you want to get official guidance from the committee on specific design items for threads.
>
> These sound good, is it worth adding new issues for them, or just add them to the agenda?

Please add all such items to the agenda. Issues are good as well. The goal is to be totally open, and help attendees prepare. No surprises should come out of the meeting, and people need some time to think things through.

>> Further, you want to ask about open work items for which you need guidance. You have 2 in there, I think there are more?
>
> I took a quick look over the design issues and couldn't find any. Are there any you're thinking of in particular?

None in particular. If there aren't any then your expectation should be to have no further work after this meeting.

jfbastien:
FYI @lars-t-hansen it would be good to get feedback, if any.

binji:
Thanks, I've added the topics you've mentioned.

2017/CG-05.md
```diff
@@ -51,9 +51,13 @@ Please register before the event [here](https://goo.gl/forms/LQVhuRuI22lK8Umw1).
        * [Proposal](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md)
        * Brief overview of the proposal
        * Topics for committee feedback
+         - [Matching SharedArrayBuffer/Atomics for v1](https://github.com/WebAssembly/threads/issues/5)
+         - [Blocking main thread in web embedding](https://github.com/WebAssembly/threads/issues/6)
+         - [float atomics, i64 atomics & i64.wait](https://github.com/WebAssembly/threads/issues/7)
```
jfbastien: Poll these separately. It's OK for the issue, but for polling you want people to be able to express themselves differently here.

jfbastien:
> Thanks, I've added the topics you've mentioned.

I editited the post and added more, am done now 😛

jfbastien:
Missing:

* Is the set of proposed opcodes sufficient?
* Is the encoding scheme good?

Oh also:

* Sequentially-consistent and non-atomic only for now?

binji:
Thanks, I think I have them all covered now. :-)

jfbastien:
I'd also ask about encoding scheme. The fact that this is the first use of prefix is something we'd want to settle, and which prefix / how we choose them later (we'll revisit for SIMD).

Otherwise this is a good start, we can add other items if needed.

binji:
> I'd also ask about encoding scheme.

I added that one too, see https://github.com/WebAssembly/threads/issues/12.
Maybe you missed it since I called it "Encoding Scheme" instead of "Is the encoding scheme good?"

binji:
Merging, for now. Like you said, we can always add more! :-)

DLehenbauer:
I couldn't get away from Redmond for this week's meeting, but I would appreciate the opportunity to champion Midori-like safety for wasm (with help from @leifko).

jfbastien:
@DLehenbauer can you open a separate PR, adding a bullet point for this and a timebox (30 min?) after the threads discussion?

PR #11, opened 2017-05-25
Update CG-05.md

dtig:
Update the agenda to include new SIMD items.

### DIFF
--- a/2017/CG-05.md
+++ b/2017/CG-05.md
@@ -555,7 +555,19 @@ Tracking issue: [Consider adding Horizontal Add #20](https://github.com/WebAssem
 
 > Include `v*.addHoriz` instructions.
 
-Packed horizontal additions are natively supported on ARMv8/SSE3. These would be useful for complex multiplications, and in the absence of the opcodes, would need to be a combination of shifts and adds. The issue proposes adding horizontal addition for f32x4, i32x4, i16x8, with a potential addendum for 64x2 types. 
+Packed horizontal additions are natively supported on ARMv8/SSE3. These would
+be useful for complex multiplications, and in the absence of the opcodes,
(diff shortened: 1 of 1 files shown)


jfbastien:
Should we only talk about horizontal addition, or should we talk about horizontal operations in general? And which types are we talking about?

I'd like to have more details here, including which ISAs support what operations.

dtig:
I've added more details from the tracking issue. We should talk about horizontal addition, as horizontal operations in general do not have consistent ISA support. 

SSE3/ARMv8 natively support horizontal addition for f32x4, i32x4, and i16x8, @stoklund proposed including f64x2 as well. Horizontal adds for i8x16 are not natively supported on the x86 ISA, but can be implemented using PSADBW.

jfbastien:
But why just add though?

dtig:
We've had some feedback that these would be useful to have for performance with complex multiplications, other operations do not have consistent ISA support (Ex. Horizontal subtracts on ARM). I can edit the agenda item to discuss horizontal operations in general if you think that would be more useful.

jfbastien:
What I meant was: the poll only mentions add, and it isn't clear why just this one. Having the extra context would be useful, otherwise we'll discuss the poll itself.

Separately, it may be worth having another poll which asks: do we want horizontal operations for other instructions? The presumption you have is that the answer is "no". That's fine! Polling for a "no" vote is a good way to ensure what you believe is what the committee also believes. Maybe just raising the poll will bring in new information which will change your opinion. Maybe people will offer guidance on what would convince them to say "yes" later (e.g. ISA support, perf numbers, specific usecases, etc).

dtig:
I see what you mean, thanks for clarifying! I've updated the pull request.

PR #14, opened 2017-05-26
Add Poll: guarantee 8- and 16-bit lock-free atomics?

binji:
Add poll about guaranteeing smaller lock-free atomics


jfbastien:
Separately, could you discuss and poll double-pointer-width? I expect we won't do it, but would like to discuss it and have it as a tentative thing we do later.

binji:
Just for clarification, do you mean 64-bit (double wasm32 pointer width) or 64-bit/128-bit (double host pointer width)?

jfbastien:
wasm32 would be 64-bit, wasm64 would be 128-bit.

PR #23, opened 2017-07-03
Add SIMD agenda items for the July 6th call.

stoklund:
Summary of the changes made to the proposal since the May CG meeting.


jfbastien:
True, we'll have to keep it on track!

PR #25, opened 2017-07-06
Adding a document describing the phases through which a proposal will pass.

flagxor:


process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+However, if issues are uncovered or consensus devolves,
+proposals should back up to the appropriate prior step.
+ 
+## 0. Feature Proposal [CG]
+
+Entry requirements:
+  * A CG member has an idea.
+ 
+During this phase a new repository forking the spec repo is created (CG members
+can request this). PRs + Issues iterate on the design of the feature. If human
```
rossberg-chromium: Nit: drop "human readable" (or replace with something else), since it implicates that the spec is not readable. ;)

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+readable design repo style .md files are useful, these are added to describe
+the feature as it's being designed. A tracking issue is created in the design
```
rossberg-chromium: What is the role of the tracking issue?

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+The feature is implemented in toolchains.
+ 
+## 3. Standardize the Feature [WG]
+
+Entry requirements:
+   * Two or more Web VMs implement the feature.
+   * At least one toolchain implements the feature.
+   * The formalisation and the OCaml reference interpreter are usually updated
+     (though these two can be done as part of step 3 at the WG chair's discretion).
+     CG + general consensus has continued to this point (as throughout).
```
jfbastien: Missing bullet here.

jfbastien: I'm not sure I understand what that means? I think what's missing is "Community Group is the sole venue where substantial work can occur." This should be stated at the beginning to clarify that iteration on the major design points of a feature are iterated on in the CG, but integrating it in the rest of the spec and getting the details right happen in the WG.

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+ 
+At this point the feature is fully handed off to the WG.
+During this phase, WG members discuss the feature, consider edge cases, and
+work towards consensus that the feature is now complete (kicking it back to an
+earlier stage in the CG if issues are uncovered).
```
jfbastien: I'd rephrase to: Unless substantial work is deemed required (in which case the feature is sent back to the Community Group).

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+Web VMs (browsers) implement the feature.
+The forked repo is updated to include revisions to the formalisation.
+The forked repo spec is updated to include implementation of the feature in the
+reference interpreter (OCaml).
+The feature is implemented in toolchains.
+ 
+## 3. Standardize the Feature [WG]
+
+Entry requirements:
+   * Two or more Web VMs implement the feature.
```
jfbastien: I think we also need some signal from others that there's interest in implementing. I think it's another one of those "the Chair decides what looks like consensus, but if people consistently feel like the Chair ignores them then things will break down".

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+## 4. The Feature is Standardized [WG]
+
+Entry requirements:
+   * Consensus is reached in amongst WG members that the feature is complete.
+ 
+When Working Group consensus is reached (online), editors can merge the feature
+into master on the spec repo.
+The W3C snapshots (for REC) are made at a regular cadence (in a W3C repo), used
+to stamp official version. Matching tags are added in the github spec repo.
+The feature is enabled on stable by various browsers, converging on all.
```
jfbastien: Is this worth mentioning? People can just ship stuff if they want, so that clause doesn't really mean much.

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+They are owned in common by the WG + CG. The WG used the
+github.com/WebAssembly/spec repo to iterate on finalized proposals for hand-off
+to the W3C's snapshotted copy of the spec. The other repos are primarily
+overseen by the CG.
+
+#### What about licenses?
+
+The spec and spec forks will move to a W3C Documents and Software License.
+Other related repos for prototypes + tools will continue under their respective
+licenses.
```
jfbastien: Point to the actual license files.

process/phases.md
flagxor: done

process/phases.md
flagxor: Added link.

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+In general, the process moves forward through a series of numbered phases.
+However, if issues are uncovered or consensus devolves,
+proposals should back up to the appropriate prior step.
+ 
+## 0. Feature Proposal [CG]
+
+Entry requirements:
+  * A CG member has an idea.
+ 
+During this phase a new repository forking the spec repo is created (CG members
```
flagxor: I think we have to a degree assume goodwill from the champion (save genuine spam).
I.e. I would think that the bar should be pretty low if a CG member says they want a repo to develop a proposal in?
Perhaps some verbiage about no objection about confusion / spam / duplication?

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+ 
+## 2. Implementation Phase [CG/WG]
+
+Entry requirements:
+   * Test suite has been updated to cover the feature in its forked repo.
+   * The test suite should run against some implementation, though it need not be
+     the reference interpreter.
+   * *Formal notation in the spec need not be updated.*
+ 
+During this phase, the following proceeds in parallel:
```
flagxor: done

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+ 
+During this phase a new repository forking the spec repo is created (CG members
+can request this). PRs + Issues iterate on the design of the feature. If human
+readable design repo style .md files are useful, these are added to describe
+the feature as it's being designed. A tracking issue is created in the design
+repo.
+ 
+If relevant to demonstrate the viability of a feature, prototype
+implementations of the feature are implemented in interested engines (possibly
+on a branch).
```
flagxor: So engine struck me as more clear, as embedder sounds relevant if the feature can wholely be done outside the core engine, but odd in reference to some part of the core format?
See below on the other variation on this.

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+## 2. Implementation Phase [CG/WG]
+
+Entry requirements:
+   * Test suite has been updated to cover the feature in its forked repo.
+   * The test suite should run against some implementation, though it need not be
+     the reference interpreter.
+   * *Formal notation in the spec need not be updated.*
+ 
+During this phase, the following proceeds in parallel:
+Web VMs (browsers) implement the feature.
```
flagxor: So my recollection was the the "WebVM" phrasing was somewhat intentional. I.e. there was an assumption here that implementations deployed to the Web ought to carry special weight.
I can see arguments for or against this, but maybe more folks should chime in?

process/phases.md
flagxor: So the last bullet was meant to cover the consensus issue (it was missing the bullet tag). I've also added a parenthetical regarding the chair's role in this.
I think the point of the two WebVM bar was that in addition to consensus, there should be track record with the feature working on the Web, given the Web's higher bar for compatibility etc., as we want to make sure the feature will eventually be universal.
Again, curious what others think on this point?

process/phases.md
flagxor: done

process/phases.md
flagxor: reworded similarly.

process/phases.md
flagxor: Domenic and others brought this one up, and my understanding is that these tags + versions carry special legal weight in terms of patent exclusion periods etc. (But I am not a lawyer).

process/phases.md
flagxor: done

process/phases.md
```diff
@@ -0,0 +1,150 @@
...
+ 
+If relevant to demonstrate the viability of a feature, prototype
+implementations of the feature are implemented in interested engines (possibly
+on a branch).
+The CG will attempt to reach broad consensus.
+ 
+## 1. Proposed Spec Text Available [CG/WG]
+
+Entry requirements:
+   * Full proposed english spec text available in a forked repo around which a
```
flagxor: done

flagxor:
PTAL

process/phases.md
jfbastien: It's not about goodwill.  People who read this document coming in will think "hmm, I have to fork a thing... eh, don't care enough".

process/phases.md
jfbastien: "Embedder" is what we use elsewhere, so it's clear in the wider WebAssembly context.

process/phases.md
jfbastien: I think that's true when we get to the WG, but at the CG level not necessarily.

process/phases.md
jfbastien: That seems unrelated to "enabled on stable by various browsers", no?

jfbastien:
I replied to a bunch of comments. Not sure GitHub shows all the replies unless you expand "outdated" comments.

process/phases.md
```diff
@@ -0,0 +1,150 @@
+# WebAssembly W3C Process
+
+This file describes how feature proposals will progress through
+the standardization process.
```
rossberg-chromium: I think we should require a proposal repo for everything that adds or changes a "feature". Criterions for that could be:

* Adds an opcode or other pieces of abstract syntax
* Changes or extends the binary format
* Changes or extends the text format
* Requires adding or changing tests
* Requires extending or changing the reference interpreter.

Any of these implies that all conforming engines are forced to be changed, which justifies going through a bit of process.

A "fix" on the other hand is something that e.g. is merely a clarification, and does not necessarily require implementations to change. That's perfectly doable as a simple PR.

jfbastien:
Ping? There are still some (now hidden) comments which need to be addressed.

process/phases.md
flagxor: Adopted those bullets.

process/phases.md
flagxor: Ah, wrong line. Yes, dropped.

process/phases.md
flagxor: Ok, added a new step zero and incremented.

process/phases.md
flagxor: Ok, done.

process/phases.md
flagxor: Ok, changed here.

process/phases.md
flagxor: Done.

flagxor:
PTAL

lukewagner:
Could we perhaps identify a clear point in the process where the CG (and/or WG?) gives a "green light" for browsers to start shipping (on-by-default) their implementation of a feature?  This would seem to be somewhere in the middle of Step 4.  I was also thinking it might be useful if this event produced some sort of linkable artifact, analogous to the [MVP email](https://lists.w3.org/Archives/Public/public-webassembly/2017Feb/0002.html).

jfbastien:
If we do this "green-light" thing, we need a phase when we say "do ship if you want, but it may break", and then we need to specify when we intend to keep compatibility as much as possible (I assume that's when it's in the spec).

lukewagner:
Ah, that's slightly different than what we messaged with the MVP email (which was that minor stuff might change, but only in minor ways that shouldn't break content in the wild).  That's sortof what green light means, at least to me.

jfbastien:
I think we might want to do discuss this addition in a video call, to be sure that everyone is on board with it. Agreed that the distinction you make from what I said is important.

Regardless of which way we go, the last bit I propose (when is it set in stone, except for *force majeure*) is still something I'd like in there.

@flagxor do you think this would be ready to discuss at next Tuesday's meeting? I think this entire document needs to be more or less ready to go by then.

flagxor:
Reluctant to map to "recommended ship" until we've discussed.
Would a WG discussion on the topic be useful?
Any other changes. Perhaps we could checkpoint this if I added a DRAFT tag on top?

jfbastien:
I think this should be a CG discussion, since it discusses how the CG's work is used. Can you add it to the agenda? Fine by me to commit with DRAFT notice, and pointer to the scheduled discussion.

flagxor:
sgtm

PR #26, opened 2017-07-06
Update following the CG video call

jfbastien:


jfbastien:
I'll merge this since we just discussed it on the call. Please send follow-ups to update as discussed.

PR #28, opened 2017-07-08
Add registration form

jfbastien:


jfbastien:
This was broken, I'm adding a new one instead.

PR #35, opened 2017-07-14
Drop several simd polls.

flagxor:


jfbastien:
Will you still discuss experience with WebP?

PR #41, opened 2017-07-17
Add poll about float timeout for wait instruction

binji:


binji:
Off-line communication w/ JF -- should be ok to add.

PR #51, opened 2017-08-01
Add agenda items for CSP and WPT

titzer:
Adds agenda items for CSP and WPT.


jfbastien:
Please make sure we have concrete proposals for both by next week. Once we do, can you also add links to both issue?

PR #52, opened 2017-08-01
Add modules presentation agenda item

domenic:
I plan to work on a set of slides over the course of this week, and will update the agenda with them when they are ready. Not sure whether this agenda item should be accepted before or after that.


jfbastien:
We can post the slides later since this is an intro presentation, getting a conversation started.

PR #64, opened 2017-08-28
Add future agenda item for data segment init

binji:


jfbastien:
Kinda weird to add to the previous agenda, but sure.

binji:
Well, we don't have the next agenda yet. :-)

PR #66, opened 2017-09-06
Adding a WG meeting.

flagxor:
Planning to announce a kickoff meeting, but want to confirm key folks can attend.

### DIFF
--- a/2017/WG-09-12.md
+++ b/2017/WG-09-12.md
@@ -1,0 +1,80 @@
+![WebAssembly logo](/images/WebAssembly.png)
+
+## Agenda for the September 14th video call of WebAssembly's Working Group
+
+- **Host**: Google Hangouts
+- **Dates**: Thursday September 14th, 2017
+- **Times**: 7:30am–8:30am Pacific Time
+- **Location**: *Brad will email Google Hangouts link to WG members prior to the meeting*
+- **Contact**:
+    - Name: Brad Nelson
+    - Email: bradnelson@google.com
+
+### Registration
+
+None required if you've attended before. Email Brad Nelson to sign up if it's
+your first time. The meeting is open to WG members only (we'll likely broaden
+this in future meetings).
+
+## Logistics
+
+The meeting will be a Google Hangouts call.
+
+## Agenda items
+
+1. Opening, welcome and roll call
+    1. Opening of the meeting
+    1. Introduction of attendees
+1. Find volunteers for note taking (chair to volunteer).
+1. Adoption of the agenda
+1. Proposals and discussions
+    1. Openness (Brad Nelson)
+       1. Discussion on how open the WG process should be.
+          * We're obviously bound by general W3C:
+          * [W3C Openness](https://www.w3.org/wiki/OpenAndTransparentW3C/Group_Openness)
+          * [Membership](https://www.w3.org/Consortium/membership-faq.html)
+       1. POLL: We will open up WG meetings to all CG members (for observation).
+    1. Mailing list
+       1. POLL: We should share the Community Group mailing lists:
+          * public-webassembly@w3.org  ( not auto-signup )
+          * public-webassembly-announce@w3.org
+    1. Scope of v.1 specification (Brad Nelson)
+       1. Discussion on which things should be included in the initial
+          specific, possibilities include:
+          1. [WebAssembly core spec](https://github.com/WebAssembly/spec)
+             1. Core format
+             1. Reference interpreter
+          1. Test suite
+          1. [JavaScript embedding](https://github.com/WebAssembly/design/blob/master/JS.md)
+             * [Draft spec](https://littledan.github.io/spec/document/JS.html)
+          1. [Web embedding](https://github.com/WebAssembly/design/blob/master/Web.md)
+       1. POLL: We should limit v.1 to core specification only.
+          This excludes: the JavaScript embedding,
+          Web embedding, reference interpreter, and test suite.
+    1. Adoption of CG specification (Andreas Rossberg)
+       1. Discussion of state of the core specification.
+       1. POLL: We should officially tag and adopt a revision of the spec as
+          the WG input.
+       1. Discussion on the process of adopting a "First Public Working Draft" for WebAssembly v.1.
+    1. Future meetings
+       1. Discussion of timing for future meetings.
+1. Closure
+
+## Agenda items for future meetings
+
+None.
+
+### Schedule constraints
+
+None.
+
+## Dates and locations of future meetings
+
+| Dates                    | Location          | Host       |
+|--------------------------|-------------------|------------|
+| 2017-11-01 to 2017-11-02 | Santa Clara, CA   | Intel      |
+| 2017-11-06 to 2017-11-07 | Burlingame, CA    | TPAC       |
+
+## Meeting notes
+
+### Roll call


2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
...
+1. Adoption of the agenda
+1. Proposals and discussions
+    1. Openness (Brad Nelson)
+       1. Discussion on how open the WG process should be.
+       1. POLL: We comfortable with opening WG meetings to all CG members (for observation).
+       1. POLL: We comfortable opening WG meetings to the public (for observation).
+    1. Scope of v.1 specification (Brad Nelson)
+       1. Discussion on which things should be included in the initial
+          specific, possibilities include:
+          1. [Binary format spec](https://github.com/WebAssembly/spec)
```
rossberg-chromium: Nit: "binary format spec" sounds weird, since that would suggest that it only defines the format, no semantics. It even contains the text format. I suggest calling it "Wasm core spec".

2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
+![WebAssembly logo](/images/WebAssembly.png)
+
+## Agenda for the September 12th video call of WebAssembly's Working Group
+
+- **Host**: Google Hangouts
+- **Dates**: Tuesday September 12th, 2017
+- **Times**: 9:00am–10:00am Pacific Time
+- **Location**: *same Google Hangouts as previous CG video meetings*
```
jfbastien: Seems odd to reuse the CG call ID if the intent is to have the first meeting with WG members only, no?

2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
...
+## Agenda items
+
+1. Opening, welcome and roll call
+    1. Opening of the meeting
+    1. Introduction of attendees
+1. Find volunteers for note taking.
+1. Adoption of the agenda
+1. Proposals and discussions
+    1. Openness (Brad Nelson)
+       1. Discussion on how open the WG process should be.
```
jfbastien: Can you link to W3C guidance on WG openness here? I'd like to understand what's up for discussion versus what's set in stone.

2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
...
+       1. POLL: We comfortable opening WG meetings to the public (for observation).
+    1. Scope of v.1 specification (Brad Nelson)
+       1. Discussion on which things should be included in the initial
+          specific, possibilities include:
+          1. [Binary format spec](https://github.com/WebAssembly/spec)
+             1. Core format
+             1. Reference interpreter
+             1. Test suite
+          1. [JavaScript embedding](https://github.com/WebAssembly/design/blob/master/JS.md)
+          1. [Web embedding](https://github.com/WebAssembly/design/blob/master/Web.md)
```
jfbastien: Indentation of the above two is off?

2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
...
+       1. POLL: We should limit v.1 to a binary format specification only.
+    1. Adoption of CG specification (Andreas Rossberg)
+       1. Discussion of state of the binary specification.
+       1. POLL: We should officially tag and adopt a revision of the spec as
+          the WG input.
+       1. POLL: We should declare this version the "First Public Working Draft".
```
jfbastien: Is this poll formally adopting it, or saying that we should perform work towards adopting it?

2017/WG-09-12.md
MikeHolman: I agree... would probably make more sense to have a different ID

flagxor: Ok, instead I'll invite WG members that have actually joined.

2017/WG-09-12.md
flagxor: Meant those two to be at same level with spec. Though I suppose for tests it's mixed in, that what you mean?

2017/WG-09-12.md
flagxor: Meant the former, though not sure how the poll will go (and likely we might not even get to that bullet). Also planning to seek some guidance from Eric + others.

2017/WG-09-12.md
flagxor: Asking for some more on this.
We have the option for instance to keep the meetings closed. I don't think we want to do that, in fact I'd rather it be wide open, but I wanted to have that be an explicit decision.

jfbastien: What do you mean "asking for more info"? I think this section should link to the specific W3C docs.

2017/WG-09-12.md
jfbastien: screenshot 2017-09-06 at 8 50 15 pm

Test suite is more indented than the following two items.

2017/WG-09-12.md
jfbastien: I don't think we can formally adopt it as-is, since we haven't been told that it was done, or that we should review it until just now.

2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
...
+          specific, possibilities include:
+          1. [Wasm core spec](https://github.com/WebAssembly/spec)
+             1. Core format
+             1. Reference interpreter
+             1. Test suite
+          1. [JavaScript embedding](https://github.com/WebAssembly/design/blob/master/JS.md)
+          1. [Web embedding](https://github.com/WebAssembly/design/blob/master/Web.md)
+       1. POLL: We should limit v.1 to a binary format specification only.
+    1. Adoption of CG specification (Andreas Rossberg)
+       1. Discussion of state of the binary specification.
```
rossberg-chromium: Nit: And here

2017/WG-09-12.md
```diff
@@ -0,0 +1,71 @@
...
+    1. Scope of v.1 specification (Brad Nelson)
+       1. Discussion on which things should be included in the initial
+          specific, possibilities include:
+          1. [Wasm core spec](https://github.com/WebAssembly/spec)
+             1. Core format
+             1. Reference interpreter
+             1. Test suite
+          1. [JavaScript embedding](https://github.com/WebAssembly/design/blob/master/JS.md)
+          1. [Web embedding](https://github.com/WebAssembly/design/blob/master/Web.md)
+       1. POLL: We should limit v.1 to a binary format specification only.
```
rossberg-chromium: Nit: s/binary format/core/ as well

flagxor: done

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: Sorry meant asking Eric.
In any event, have added links and rephrased the poll slightly. It sounds like the main constraint is that we need to attest that artifact produced by the WG come from folks bound by the proper agreements.

flagxor:
PTAL

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
+![WebAssembly logo](/images/WebAssembly.png)
+
+## Agenda for the September 12th video call of WebAssembly's Working Group
```
jfbastien: Date is wrong?

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+
+- **Host**: Google Hangouts
+- **Dates**: Tuesday September 14th, 2017
```
jfbastien: Thursday.

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+## Logistics
+
+The meeting will be a Google Hangouts call.
+
+## Agenda items
+
+1. Opening, welcome and roll call
+    1. Opening of the meeting
+    1. Introduction of attendees
+1. Find volunteers for note taking.
```
jfbastien: (chair to volunteer)

😉

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+1. Adoption of the agenda
+1. Proposals and discussions
+    1. Openness (Brad Nelson)
+       1. Discussion on how open the WG process should be.
+          * We're obviously bound by general W3C:
+          * [W3C Openness](https://www.w3.org/wiki/OpenAndTransparentW3C/Group_Openness)
+          * [Membership](https://www.w3.org/Consortium/membership-faq.html)
+       1. POLL: We will open up WG meetings to all CG members (for observation).
+       1. POLL: We will be generous in inviting non-WG members as invited experts when appropriate to WG meetings.
```
jfbastien: "be generous" sounds like we're assholes. They're being generous by calling in! I'd phrase it more as doing the best we can to get the right experts on the call, even if they're not WG members.

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+    1. Mailing list
+       1. POLL: We should share the Community Group mailing lists:
+          * public-webassembly@w3.org
+          * public-webassembly-announce@w3.org
```
jfbastien: Note that this one is moderated, and low-traffic. BTW I believe that you must email that list with the meeting details, and then you must email it with notes afterwards. At least that's what I do because IIRC that's one of the W3C requirements (I think?).

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+    1. Openness (Brad Nelson)
+       1. Discussion on how open the WG process should be.
+          * We're obviously bound by general W3C:
+          * [W3C Openness](https://www.w3.org/wiki/OpenAndTransparentW3C/Group_Openness)
+          * [Membership](https://www.w3.org/Consortium/membership-faq.html)
+       1. POLL: We will open up WG meetings to all CG members (for observation).
+       1. POLL: We will be generous in inviting non-WG members as invited experts when appropriate to WG meetings.
+    1. Mailing list
+       1. POLL: We should share the Community Group mailing lists:
+          * public-webassembly@w3.org
```
jfbastien: Note that this one isn't auto-sign-up, people need to add themselves.

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+          * public-webassembly-announce@w3.org
+    1. Scope of v.1 specification (Brad Nelson)
+       1. Discussion on which things should be included in the initial
+          specific, possibilities include:
+          1. [Wasm core spec](https://github.com/WebAssembly/spec)
```
jfbastien: I'd rather spell out "WebAssembly" everywhere instead of "wasm".

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+             1. Core format
+             1. Reference interpreter
+          1. Test suite
+          1. [JavaScript embedding](https://github.com/WebAssembly/design/blob/master/JS.md)
+          1. [Web embedding](https://github.com/WebAssembly/design/blob/master/Web.md)
```
jfbastien: Link to @littledan's initial work at making these spec-ish?

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+       1. POLL: We should limit v.1 to core specification only.
```
jfbastien: Implicit in this question is we're leaving things out. You have a list above of what's in, could you also list what's out so that people know what we're excluding when they vote on the poll?

2017/WG-09-12.md
```diff
@@ -0,0 +1,78 @@
...
+    1. Adoption of CG specification (Andreas Rossberg)
+       1. Discussion of state of the core specification.
+       1. POLL: We should officially tag and adopt a revision of the spec as
+          the WG input.
+       1. POLL: We should formally declare this version the WG's "First Public Working Draft" for WebAssembly v.1.
```
jfbastien: I'm still inclined to vote against this: we need to figure out how we'll review and adopt the work. Phrased like this it sounds like an outright adoption without review.

2017/WG-09-12.md
littledan: FWIW current draft here https://littledan.github.io/spec/document/JS.html, but it is missing conversion for the most interesting parts.

2017/WG-09-12.md
ericprud: @jfbastien , I wonder if blanket adoption would be appropriate. Adoption effectively means that you copy the contents into a new page which is managed by the WG. Upsides:

* Get started quickly.
* Immediately have a document against which the WG can post issues and PRs.

downside:

* Strongly biases WG toward the adopted document.
* Delays pressure on WG to review the doc.

There's a lot of pressure for the WG to review the doc after the editors claim it's ready to be published as a First Public Working Draft. Is it practical push the WG to review before adoption? Is it likely to change the outcome? My guess is that we'll adopt it eventually and our process for working on it will be better-defined once we adopt it.

jfbastien: Maybe #25 needs an update to clarify. In any case, this question is ill-phrased and needs some clarification ASAP. If you proposal is outright adoption I want to know what it means, and I want sufficient time to consult internal folks (and I expect other companies want them same).

Without sufficient time and clarity the only sane vote is Strongly Against. The usual process is to revisit only when there's new data, which a clarification would be.

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: Good point dropped.

2017/WG-09-12.md
flagxor: noted.

2017/WG-09-12.md
flagxor: yep

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: added.

2017/WG-09-12.md
flagxor: done

2017/WG-09-12.md
flagxor: I actually wasn't sure we wanted to do that immediately, but put it out to be provocative.
I've reduced it to a discussion topic.

flagxor:
I've addressed outstanding items, and am going to merge and get this out so we can announce with proper notice.

PR #68, opened 2017-09-13
Add CG September 19th video call

jfbastien:


jfbastien:
@flagxor I put you on the hook for a summary of the WG call.

@titzer can you please expand on the item you suggested at the last meeting?

PR #76, opened 2017-10-03
Add Oct 17th VC

jfbastien:


jfbastien:
@rossberg-chromium and @titzer please confirm that the items in your name are OK, and add details if so. We can punt to a future meeting if not.

rossberg-chromium:
LGTM

PR #80, opened 2017-10-12
Fill out agenda item for WPT

titzer:
This fills out the agenda item for the WPT issue.

### DIFF
--- a/2017/CG-10-17.md
+++ b/2017/CG-10-17.md
@@ -30,6 +30,7 @@ The meeting will be a Google Hangouts call.
     1. [Versioning of external standard dependencies](https://github.com/WebAssembly/spec/issues/566) (Andreas Rossberg).
     1. Web platform test repository, discussion, and poll (Ben Titzer).
        1. Background: The [Web Platform Tests repository](https://github.com/w3c/web-platform-tests) is W3C-coordinated attempt to build a cross-browser testsuite for the Web-platform stack. It incorporates tests from many sources and offers, for example, a two-way sync feature between the Chromium repository and WPT in an effect to make it much easier to expand the WPT tests. The WPT effort would like to incorporate as many WASM tests as possible and run them in a cross-browser fashion, and make it easy to add tests.
+       The issue is complicated by a build step that is necessary, since WASM spec tests are almost all written using the text format, which requires translation to binaries and JS files for running in engines that have a JS embedding.
        1. Discussion of proposals
            1. Strawperson A: WPT becomes the source of truth
               1. Build a 2-way sync between V8 (other engines) and WPT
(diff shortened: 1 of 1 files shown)


2017/CG-10-17.md
```diff
@@ -29,7 +29,19 @@ The meeting will be a Google Hangouts call.
 1. Proposals and discussions
     1. [Versioning of external standard dependencies](https://github.com/WebAssembly/spec/issues/566) (Andreas Rossberg).
     1. Web platform test repository, discussion, and poll (Ben Titzer).
-       1. Details to be added before meeting.
+       1. Background: The [Web Platform Tests repository](https://github.com/w3c/web-platform-tests) is W3C-coordinated attempt to build a cross-browser testsuite for the Web-platform stack. It incorporates tests from many sources and offers, for example, a two-way sync feature between the Chromium repository and WPT in an effect to make it much easier to expand the WPT tests. The WPT effort would like to incorporate as many WASM tests as possible and run them in a cross-browser fashion, and make it easy to add tests.
```
jfbastien: "in an effort"?

jfbastien:
Could you also clarify if you're talking about core tests, and / or Web / JS embedding tests? I think @lukewagner had opinions on that.

titzer:
Per offline conversation, @flagxor will add a link to a doc with more background.

foolip:
@domenic pointed me here for the web-platform-tests agenda item. Some thoughts:

To clarify "the WPT effort would like to incorporate as many WASM tests as possible": while tests for any web-exposed features are welcome in WPT, it's not a goal to subsume all other test suites, what matters is that shared test suites exist, are well maintained, and that there's good infrastructure around them.

I have no involvement in wasm, but I think these are the important questions:
 1. Would implementers like to add/change shared tests together with their code changes? (IMHO, this is very important, as otherwise using shared tests always comes with an overhead. See [this email](https://groups.google.com/a/chromium.org/d/msg/ecosystem-infra/yDAnUCUqFcw/sk5gxzjbBgAJ) for how much of a difference it seems to make.)
 1. If yes, how do you enable two-way sync between the repos involved?
 1. How do you get an always-fresh view of the state of the test suite, by which implementers can prioritize their work?

If merging into wpt and using wpt.fyi is the best answer, that'd be OK, and I don't anticipate you'd meet with any resistance. The build step could perhaps happen at runtime, looks like it's a [python script](https://github.com/WebAssembly/spec/blob/master/test/core/run.py) already.

As for a one-way sync in order to get results into wpt.fyi, someone would need to keep that always running and make sure that the special directory can't be changed in wpt and that people can immediately discover why. A separate wasm.fyi would feel cleaner, but that's a lot of work.

It would be nice if the outcome is all wasm tests in one place, not split over [spec/test](https://github.com/WebAssembly/spec/tree/master/test) and [wpt/wasm](https://github.com/w3c/web-platform-tests/tree/master/wasm) as currently.

jfbastien:
@foolip I recommend commenting here:  https://github.com/WebAssembly/spec/issues/529
I don't think this PR is as permanent as the above issue.

foolip:
@jfbastien, thanks, done!

PR #123, opened 2017-11-23
Add agenda items for Nov 28

lars-t-hansen:
Two open issues on the Globals design and one on the Threads design.


binji:
If nobody minds, I'll just fix this since I'm assuming it's midnight wear Lars is and the meeting is tomorrow.

chicoxyzzy:
It seems like this PR could be closed because #124 landed

PR #173, opened 2018-02-21
Add link to slides for CSP discussion

eholk:


jfbastien:
I feel like this discussion is way too focused on a single thing, and is totally missing a big picture. IMO we can't really move one sub-part of CSP / SRI forward without having the whole picture. It would be helpful to have a table with all WebAssembly APIs (rather, each usage of each API), and how each CSP setting (none, unsafe-eval, etc) affects each API. Ditto for SRI.

Put another way, how does a developers allow / disallow certain behaviors. I'd focus on what each setting allows and prevents, and why that's a useful security feature. Specifically, in the context of WebAssembly, what XSS would be possible without a particular CSP setting, which is now impossible with that setting.

We probably have this in a few locations, but to move forward you really need to pull all of this together so we all have a coherent picture.

flagxor:
I've added something like the full table to this:
https://github.com/WebAssembly/content-security-policy/blob/master/proposals/CSP.md

Eric, can you revise to the extent this has changed?

PR #194, opened 2018-03-26
Add agenda item about opcode numbering

binji:


jfbastien:
Could you summarize the current state of decision (from the in-person meeting), so it's easy to extrapolate a sensible policy?

binji:
OK, at the meeting you mean, right? I can do that.

jfbastien:
Yeah, would be good to write it down in the agenda so people come in prepared.

PR #204, opened 2018-04-06
Add poll about WebAssembly.Global default value

binji:


binji:
This should be a quick topic, added it to Wednesday since it seemed like a lighter day, but we can probably squeeze it in anywhere.

binji:
@jfbastien I fixed the conflict, does this seem like an OK time or should we move it?

jfbastien:
You mean `+0.0` 😄

jfbastien:
I'll reshuffle the schedule as we get closer to the date.

PR #217, opened 2018-04-10
Add link to ESM integration proposal

tschneidereit:


tschneidereit:
@jfbastien not sure if you saw this. People might not be aware of the location of the ESM integration proposal, so the link would allow them to better prepare.

PR #274, opened 2018-07-10
doc(process): include details in the attendance process

ashleygwilliams:


ashleygwilliams:
i realize now that that doc is almost exclusively about in-person meetings? i can try to update it to include online meetings. let me know what you think is best.

jfbastien:
Can you update this doc to have an in-person and a video call section?

dtig:
Merging this as it is an improvement to the current documentation.

PR #295, opened 2018-08-15
Add WG-08-29 meeting

binji:


binji:
@kmiller68 I was going to merge this with at 9am, but then I noticed that most Wednesdays the Google wasm team has an all-hands meeting at that time. That doesn't mean we can't move the WG meeting then, but it would be a bit disruptive. I'll still ask what folks think about moving slots though.

kmiller68:
I see. It's not the end of the world if it stays at 8am.

PR #317, opened 2018-10-02
Add CG-10-16 and WG-10-10

binji:
Also add agenda items that slipped from the previous CG meeting.


binji:
cc @fitzgen

fitzgen:
Thank you!

PR #327, opened 2018-10-23
Make exception handling the first item on Friday

aheejin:
As I plan to dial in, in the current state that's likely to be 11am in
Lyon and 2am in California. 10am there / 1am in Cal would be slightly
better.


binji:
Sounds good, though we have the meeting room at 8:30, so we'll probably start around 9:00 instead (12:00AM PDT)

PR #347, opened 2018-12-09
Preproposal phase update

arunetm:
Updating pre-proposal phase description to recommend issue creation in the proposal repo instead of design repo.

### DIFF
--- a/process/phases.md
+++ b/process/phases.md
@@ -29,7 +29,7 @@ Entry requirements:
 
 Pre-proposal process:
 
-  1. An issue issue is filed on the [design repository](https://github.com/WebAssembly/design/issues) to present the idea.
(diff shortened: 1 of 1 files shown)


aretmr:
@binji please review the accuracy of this update to the description of phases.

binji:
Seems like a good idea. One reason folks liked having a separate proposals repo is that it would be less noisy than following the design repo, and this could change that. But we haven't seen a huge number of new proposals, so it's probably OK.

chicoxyzzy:
IMO it's better to keep new pre-proposals in design repo.

dtig:
As it's been some time, and we've continued to use the design repo for new proposals I'm closing this PR - please feel free to reopen/follow up if necessary.

PR #380, opened 2019-03-14
markdown incorrectly placed community group link

birm:
The link currently goes to https://github.com/WebAssembly/meetings/blob/master/process/w3.org/community/webassembly, which 404s.


binji:
Thanks!

PR #382, opened 2019-03-18
Add registration form

binji:


binji:
@littledan does the form look OK? I copied from a previous meeting, not sure about what the food situation will be.

littledan:
LGTM. We're planning on providing lunch and snacks in the office, for what it's worth.

PR #392, opened 2019-04-05
Add K presentation to CG-04-16

binji:


binji:
@chicoxyzzy do you mind if we do this presentation first? It will likely take a fixed amount of time, and the discussion around inclusiveness will vary.

chicoxyzzy:
@binji sure, I think it's better to make discussion about inclusion to the end of agenda

grosu:
How much time do we have for the presentation on K?  @ehildenb will most likely give our presentation, who is also working on the WASM semantics in K: https://github.com/kframework/wasm-semantics.  He will talk about both K and the current status of WASM in K.  Hope this is OK.

binji:
> How much time do we have for the presentation on K?

The meeting is an hour long, and we have a discussion afterward. I think 20-30 minutes would probably be OK.

> Hope this is OK.

Yes, sounds good!

PR #396, opened 2019-04-09
Add an agenda item for landing WASI libc in WebAssembly/reference-sysroot

sunfishcode:


binji:
Seems OK to me as long as there doesn't end up being too much discussion. I want to make sure to leave time for the later agenda items.

PR #430, opened 2019-07-03
Add exception handling slides for CG-06

aheejin:


aheejin:
> It looks like the link is not publicly accessible at the moment. We could either copy over the slide-deck to a chromium.org account where it is easier to make public or just safe a PDF version of it. WDYT?

I transferred it to chromium.org account. Do you think making PDF is better?

mstarzinger:
> I transferred it to my chromium.org account. Do you think making PDF is better?

Cool. Thanks! I think using the chromium.org account as in the last revision is best.

rossberg:
Personally I'd love if all presentations were available in the meeting repo as PDF snapshots, since external docs might get modified or even die later. But that's not a requirement.

PR #452, opened 2019-09-26
Update CG-10-01.md

eqrion:


eqrion:
Would it be possible to merge this so that people are aware this will be discussed?

PR #470, opened 2019-12-06
Add agenda item

lars-t-hansen:


dtig:
Thanks for the PR, we do have updates on the CG meeting. The dates are February 11th, 12th in San Francisco. Will be adding a registration link to this agenda item shortly as well.

PR #489, opened 2020-01-29
Add upcoming February, March meetings

dtig:


binji:
I think we should cancel the February 4th meeting, since we'll have the f2f meeting the following week.

dtig:
Sounds good, I was going to cancel the one after on 2/18 and leave this one in, will now cancel both.

PR #507, opened 2020-02-14
Add follow-up to module sharing and encapsulation

RossTate:
I've prepared slides based on questions people asked after the discussion, but I'll be calling in from Orlando airport where I'll be taking off from shortly after the meeting, so we might have to push back the discussion to the next CG meeting. But it's worth a try!

### DIFF
--- a/2020/CG-02-18.md
+++ b/2020/CG-02-18.md
@@ -25,6 +25,7 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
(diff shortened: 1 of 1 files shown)


dtig:
@RossTate As this is currently the only item on the agenda, and it sounds like you may be in transit at the time with a tentative push to the next CG meeting anyway, I'll add this to the agenda for the next meeting and cancel this one. Please reach out if that doesn't work for you.

RossTate:
Yeah, no problem at all! I was expecting to do this next meeting anyways and was just adjusting to the schedule change. It'll be much nicer not presenting from the airport!

dtig:
Thanks, makes sense! I've merged it into the agenda for the next meeting.

PR #508, opened 2020-02-18
Add item on module sharing and encapsulation

dtig:


dtig:
Cc: @RossTate.

RossTate:
Yep, move works for me.

PR #517, opened 2020-03-04
Add notes for 03/03/2020 meeting.

dtig:


dtig:
@RossTate Could you add/link slides for your presentation on Module Sharing and Encapsulation?

PR #518, opened 2020-03-04
Add slides on roadmaps

RossTate:


dtig:
Thanks!

RossTate:
@dtig By the way, is this the best way for me to upload slides to the repo? Seems a bit circular, though that might just be the GitHub way :)

dtig:
Anyway that works is usually fine. One streamlined way to do this would be to add link in slides with the initial agenda PR if available, or ahead of the meeting so that folks attending the meeting have a chance to look at it ahead of time, or look at slides for reference during the presentation - but this is not always possible. Outside of that - commenting with a link to the slides in the original PR, or just uploading slides onto the repo works too and one of us can link them to the notes before merging. :)

PR #520, opened 2020-03-16
Defer GC subgroup topic

rossberg:
Unfortunately, I can't attend tomorrow due to a family incident, so I'm pushing the item to the next meeting.

### DIFF
--- a/2020/CG-03-17.md
+++ b/2020/CG-03-17.md
@@ -25,7 +25,6 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
-    1. Form GC subgroup?
 1. Closure
 
 ## Agenda items for future meetings


RossTate:
Given that this is purely organizational and was already deferred from the in-person meeting, could someone like @lukewagner not take this discussion over so that we can get the momentum building?

binji:
I agree that this is mostly organizational. Do we need to have any discussion more than agreeing to create the subgroup?

dtig:
Perhaps we should discuss this offline instead of at a meeting? I suspect the forming of the subgroup itself is non-contentious, but there may be open items to discuss regarding the details of the charter. Are there any objections to cancelling this meeting and discussing the details in an issue? That way we can possibly make progress without waiting the additional two weeks, and @rossberg can also chime in asynchronously when possible.

RossTate:
Hmm, I guess the question is how do we determine who wants to be involved, either in the subgroup or in hashing out whatever details need to be resolved to form a subgroup? (I do not know the subgroup process, so I have no idea what those details are.)

binji:
Generally anyone who wants to be involved can be. There is no official subgroup process, but the way we've done it in the past is to have a charter, and separate meetings for that group. In essence it's primarily a way to have more focused discussions.

RossTate:
Sounds good. I realized I should have said earlier that I have no objection; I am just wondering as a newcomer how we invite people to get involved.

binji:
> I am just wondering as a newcomer how we invite people to get involved.

All they have to do is [join the CG](https://www.w3.org/community/webassembly/), then they can be involved. With other subgroups we've had regular (usually every two weeks) meetings, I'd suggest we do the same here. See, for example, the [interface types meetings](https://github.com/WebAssembly/interface-types/issues/81) or [simd meetings](https://github.com/WebAssembly/simd/issues/202).

binji:
I've [created an issue here](https://github.com/WebAssembly/gc/issues/80) to discuss. In the meantime, let's cancel tomorrow's meeting.

dtig:
Thanks Ben, leaving a link here to the [debugging subgroup meetings](https://github.com/WebAssembly/debugging/tree/master/meetings) as well. Meeting tomorrow is now cancelled.

RossTate:
Thanks for all the links and filing the issue to get things going! I'll get reading to catch up :)

PR #522, opened 2020-03-17
Cancel 03/17/2020 meeting.

dtig:


penzn:
@dtig "Cancel meeting on ..." PRs are very handy - thank you! Some mail clients (just one that I know of to be honest 😄) struggle to update our meetings, but this makes it pretty unambiguous.

dtig:
Thanks @penzn for the feedback, we will try to do this more going forward.

PR #533, opened 2020-04-01
Add an agenda item to discuss safety in the WebAssembly C API.

sunfishcode:


RossTate:
In yesterday's meeting, it sounded like @rossberg might not be able to make this despite being interested. I'm not sure of that fact, so I just wanted to give a nudge to double-check that we don't have a scheduling mishap.

dtig:
The conclusion from yesterday's meeting was to schedule another off cadence meeting to discuss subtyping after the next CG meeting so @rossberg can attend as well - this is the regularly scheduled meeting, nothing needs to change here.

rossberg:
Thanks @RossTate, indeed, I won't be able to make the next meeting. @sunfishcode, would you mind if we pushed the C API discussion to the meeting after? Or is it super urgent?

sunfishcode:
Moving it to the meeting after works for me. I've now submitted https://github.com/WebAssembly/meetings/pull/538 to move it.

PR #535, opened 2020-04-02
Add notes for off cadence CG meeting on subtyping

dtig:


dtig:
Leaving a note here for @rossberg to add slides presented in the meeting today.

PR #536, opened 2020-04-03
Slides about subtyping

rossberg:


rossberg:
Took the liberty to land, so that I have a link for the notes PR.

dtig:
Makes sense, thanks for adding slides.

PR #539, opened 2020-04-06
Add discussion of low-level stack and control primitives

RossTate:
There is no plan to make any decisions. Rather, I'm hoping to just review what these primitives are and then have the group brainstorm how/if they might fight into WebAssembly.


bitwalker:
Definitely interested in this, thanks for adding!

PR #552, opened 2020-04-30
Add CG-05-12.md and pre-populate with an agenda item

eqrion:
The agenda item was from 03-31 but never got discussed.


dtig:
Thanks for adding.

PR #554, opened 2020-04-30
Add meeting notes

dtig:


2020/CG-04-28.md
```diff
@@ -48,4 +46,238 @@ Installation is required, see the calendar invite.
...
+ 
+KM: Is the reason why this is better... it makes code generation easier, since your loop can be dependent on what the machine supports?
+ 
+PP: Yeah, generally because you would have to.. We don’t actually combine them 
+ 
+KM You could do a loop invariant strength reduction, there are loops you can write that you don’t have to flatten it out.. 
+ 
+AR: Similar concern as KM
+ 
+JP: Consensus requirements are low: do members believe that the feature is in-scope will be workable.
```
taralx, line 271: I think this was me, not Jay?

dtig: Yes, this was - it's a case of conflicting initials for JP Sugarbroad and Jay Phelps. We usually try to disambiguate in the notes with when listing attendees if there are multiple folks with the same initials speaking at the meeting, but it's not always easy to keep track. Feel free to submit a PR to disambiguate (using JS, or whatever other way you would prefer).

binji: Yep, this most commonly happens with Deepti Gandluri and Dan Gohman too... :)

PR #555, opened 2020-05-01
Add poll for BigInt<->i64 advancement

Ms2ger:


Ms2ger:
I added this before the discussion point because it should not take an open-ended amount of time.

CC @xtuc

Ms2ger:
Sorry, didn't see #553 before.

xtuc:
I agree with @Ms2ger that the discussion should take too long.

dtig:
No issues, will merge this one and delete the other.

PR #565, opened 2020-05-25
Advance references

rossberg:


2020/CG-05-26.md
```diff
@@ -28,6 +28,8 @@ Installation is required, see the calendar invite.
     1. Should we move proposals entirely out of the design repo? (JP Sugarbroad)
     1. Poll to include .bitmask operations in the current SIMD proposal.    
        Context with benchmarking data: [PR](https://github.com/WebAssembly/simd/pull/201), [Issue](https://github.com/WebAssembly/simd/issues/131)
+    1. Poll to advance [reference types](https://github.com/WebAssembly/reference-types/) to phase 4 (Andreas Rossberg)
+    1. Poll to advance [typed (function) references](https://github.com/WebAssembly/function-references/) to phase 2 or 3 (Andreas Rossberg)
```
Ms2ger, line 32: @rossberg The phase document requires "Full proposed English spec text available". Could you point to it?

rossberg: Well, we have abandoned that requirement per discussion at the last F2F, and now only require a sufficiently complete informal description that would allow implementing the feature, as was the case with various other proposals (we should update the doc). The description in the overview's sections Language and following is complete and precise, though arguably a bit too terse to qualify.

I'm fine with deferring if that's what the CG thinks. Mainly, I want to bring recent progress on the proposal to broader attention.

PR #569, opened 2020-05-28
Add Await proposal agenda item

kripken:
Discussion in the issue [seems promising, but also big open questions](https://github.com/WebAssembly/design/issues/1345#issuecomment-635649331)
which could be discussed in detail during stage 1.


binji:
FYI: we probably will have time for this, but it might get bumped to the next meeting.

PR #571, opened 2020-06-02
[gc subgroup] Add notes for June 2nd call

binji:


binji:
As discussed in #549, I'm putting the notes for the GC subgroup meeting in a new top-level subdirectory called `gc`.

binji:
Merging for now, @rossberg @RossTate please update with your presentations.

PR #573, opened 2020-06-02
Move wasm CG/WG meeting notes to top-level main/

binji:
As discussed in issue #549.


rossberg:
Hm, `wasm` seems a bit generic, as this all is Wasm. How about `main`?

binji:
I'm OK with `main` (I mostly wanted to avoid `cg` since that's already very easily confused with `gc`).

binji:
Updated to `main`, PTAL

PR #578, opened 2020-06-08
Add ref.is_null item

rossberg:


dtig:
With the removal of the Await proposal discussion, we should have time for this in the 06/09 meeting. Merging this in with a note that this may get bumped to 06/23 in case we don't get to it.

PR #580, opened 2020-06-09
Add items for intrinsics and vote scheduling

tlively:


binji:
Why the given ordering of topics?

tlively:
I expect the intrinsic imports discussion to build naturally on the ITT discussion, but it would be fine to have it after the Discord discussion as well.

dtig:
I've opened #581 to start an offline discussion about vote scheduling so we can save some time and keep the agenda item if we need it in a couple of weeks from now.

RossTate:
Is there some prior discussion of intrinsics that would be good to read ahead of the meeting?

binji:
> Is there some prior discussion of intrinsics that would be good to read ahead of the meeting?

Not that I know of... a quick grep through the notes only shows references to SIMD intrinsic functions. My recollection from a while back was that we tried to avoid intrinsics (preferring instructions instead), but I'm not sure if that's written down anywhere.

tlively:
There are some comments about intrinsics on the await issue and the ITT issue, specifically:

 - https://github.com/WebAssembly/design/issues/1345#issuecomment-631183810
 - https://github.com/WebAssembly/design/issues/1345#issuecomment-631190167
 - https://github.com/WebAssembly/design/issues/1345#issuecomment-631203207
 - Discussion starting with https://github.com/WebAssembly/design/issues/1344#issuecomment-634256282

Since this has come up multiple times recently, I wanted to have a focused discussion about it. I will open a separate design issue for discussion.

RossTate:
Thanks for the nice summarized list! Don’t feel a need to open an issue on my account, though I do like there being an offline pre-discussion before online discussions in general.

RossTate:
@tlively, it occurs to me that this might be related to what you have in mind? https://github.com/WebAssembly/proposal-type-imports/issues/6#issuecomment-631685100

tlively:
omg I never merged this 🙃 RIP

binji:
I've added the next meetings here, so you can push it to the next one: https://github.com/WebAssembly/meetings/blob/master/main/2020/CG-07-07.md

PR #588, opened 2020-06-23
Update phases.md

taralx:
Pretty sure this is a typo. :)


taralx:
Okay, how about this? Still a 1 word change, but I feel like it captures the intent better.

oneemo95 (commented):
Duplicate of #

PR #594, opened 2020-06-23
Add discussion of identity in wasm vs JS

RossTate:


binji:
Would you mind providing a time estimate for this discussion (as mentioned in #595)? I'd like to try it out for this meeting.

RossTate:
I'm guessing 15-20. Should I put a range in and then when the meeting comes up and the items are more settled you can decide where in that range the box should be?

binji:
Yep, that sounds good to me. @tlively added a little `[15 minutes]` after his agenda items, can you do the same here?

RossTate:
Done

PR #595, opened 2020-06-23
Add time estimates

tlively:
These are my best guesses rounded to 5-minute increments, although I'm sure we could talk about the second item for much longer if we wanted to.


binji:
lgtm, but since we're time-boxing, you may want to bump it up to 20 minutes if you think it needs it.

tlively:
Sure, I might do that closer to the meeting time if the agenda isn't already full. 15 minutes seems fine for now, though 👍

PR #608, opened 2020-07-28
Add agenda items

rossberg:


RossTate:
I feel like these should be written up and discussed offline first.

rossberg:
@RossTate, done, see WebAssembly/design#1359.

RossTate:
Thanks! That's on the first point. But what about the second point? That's the one you brought up at the last meeting.

binji:
I'm assuming the second point is also a presentation? If so, I think it's OK to not have an associated issue (we haven't required that in the past).

rossberg:
Yes, the latter is more like grounds for discussion.

main/2020/CG-08-04.md
```diff
@@ -25,6 +25,10 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
+    1. Continuations & effect handlers as typed stacks ([issue](https://github.com/WebAssembly/design/issues/1359
+)) (Daniel Hillerström, Daan Leijen, Sam Lindley, Matija Pretnar, Andreas Rossberg, KC Sivaramakrishnan) [20 min]
+       - Poll: Phase 1?
```
kripken, line 30: How does phase 1 work given there is another approach being discussed - would both approaches be discussed inside a single repo, sort of like GC is doing? Or would there be separate repos?

PR #611, opened 2020-07-31
Add Lightly Typed WebAssembly

RossTate:
A quick presentation on how all the currently planned functionality for WebAssembly can be achieved with much less type complexity.


binji:
I'm a little concerned that adding this agenda item puts us right at one hour. Can we push this to the next meeting?

RossTate:
I would very much like to present it this meeting. There's a block of 35 minutes reserved for essentially one topic, which seems excessive. Alternatively, I can rehearse my presentation to make sure it fits within 5 minutes.

binji:
I assume this is related to https://github.com/WebAssembly/design/issues/1352 and https://github.com/WebAssembly/function-references/issues/27? If so, I think we'll want to have more than 5 minutes to allow for discussion. I don't agree that 35 minutes is excessive; at the last meeting we spent as much time on the stack inspection discussion.

RossTate:
It's not related to those issues.

RossTate:
Those issues are about fewer type *annotations*. This is about fewer *types*.

binji:
OK, I think we can add this to the meeting after the other agenda items, understanding that it may be bumped to the next meeting.

RossTate:
Thanks. I'll have a 5 minute version prepared.

PR #612, opened 2020-08-04
Add overflow item for stacks discussion

rossberg:


fgmccabe:
I do not feel comfortable with a poll being organized for this proposal with another proposal not having had 'air time'

binji:
I agree. This is a unique situation, having two proposals on the same topic at the same time. I think we should discuss as a group how best to handle this before we accept either as a phase 1 proposal.

PR #614, opened 2020-08-04
Add presentation for EH to support two-phase unwinding

aheejin:
This is a presentation that summarizes
WebAssembly/exception-handling#123. I think the discussion is likely to
take long, so I'm not sure how long I should book for this.


binji:
25 minutes seems OK. Can you go into more detail about the poll?

aheejin:
The poll is basically whether to make the change I propose or not. I'm not sure if we should be ready to take a poll on this change or not by the end of the presentation though, because the discussion might run overtime, or people might need more time to discuss the issue in relation to both first class stack proposals.

binji:
Sounds good, thanks!

PR #625, opened 2020-08-19
Add two-phase unwinding discussion part 2

aheejin:
In the last meeting we only covered the problem part, so I'd like to
present and discuss our proposed changes.

### DIFF
--- a/main/2020/CG-09-01.md
+++ b/main/2020/CG-09-01.md
@@ -25,6 +25,8 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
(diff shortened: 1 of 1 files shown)


binji:
Sounds good, I just want to make sure we give time for discussion in this session.

aheejin:
Still not sure 25 mins is enough for remaining presentation + discussion + poll, but I was also not sure if it was OK to hog the whole meeting like 45 mins for my agenda alone..

binji:
We already have https://github.com/WebAssembly/meetings/pull/626 too. So we should try to make 25-30 min work here, if we can. I wouldn't want it to extend out to another two weeks.

aheejin:
Increased it by 5 mins in #628.

PR #626, opened 2020-08-19
Add discussion of First-Class Stacks

RossTate:
The proposal has been in Phase 0 for 3 weeks (WebAssembly/design#1360). It seems there's broad interest in the feature, and so far no one has expressed concerns with the design, so it seems ready to discuss with the CG broadly and poll for Phase 1.


binji:
I'm OK with the discussion but I think we should hold off on the phase 1 poll. This is the first time we've had two proposals on the same topic start at the same time, and i think we should discuss as a group how best to handle this.

RossTate:
Yep, makes sense.

PR #638, opened 2020-09-10
Discussion on branch hinting

yuri91:
I would like to briefly present the idea, possible implementations, and use cases, and gather some feedback on how to proceed with an actual proposal.
If the schedule is already too tight, I can move this to the next meeting.


binji:
Thanks for adding this! Would you mind pushing this to the next meeting, since this one is already at 40 minutes?

yuri91:
Yes, no problem. I am closing this PR then

PR #644, opened 2020-09-17
Create SG-09-25-20.md

fgmccabe:
Set up next friday meeting of Stacks subgroup


binji (approved):

rossberg:
If only we were using date formats that actually sort properly... ;)

taralx:
I'm not sure why there's a -20 suffix -- it's already in the 2020 folder.

PR #651, opened 2020-09-23
Propose relaxed dead code validation (Phase 1)

conrad-watt:
Co-champions: @conrad-watt, @RossTate 

This can be bumped to the next meeting if necessary due to timing, but it's highly related to Andreas' `select` typing issue, so I'm hoping to piggy-back on his explanation of the issue (https://github.com/WebAssembly/design/issues/1379). This proposal is for the smaller version of the change described in the linked issue.

My vision would be to have an relatively uncontroversial phase 1 vote this week to create the proposal repo, and then a much longer presentation/discussion next meeting with a possible phase 2 vote (the spec changes are small - we mainly need commitment from implementers on the decode/validation changes to advance further).

Note that it's fine for Andreas' proposed typing change to land first - we're not proposing to make reference types depend on this proposal.


binji:
I think this sounds OK. I'm a little nervous about running out of time, but I think it should be a pretty straightforward poll.

PR #661, opened 2020-10-12
Add announcement for Oct 23 SOIL Seminar

RossTate:


tlively (approved):
👍

sbc100:
Was there a recording made of today's one?   (Obviously it only just happened so maybe a delay but I'd love to see it if/when its available).

penzn:
I'd love to see that too - went to add the meeting to calendar only to find out I was an hour late :(

RossTate:
The recording can be found here: http://soil-initiative.org/seminar/#20201012

PR #690, opened 2020-12-28
Update on Branch Hinting proposal

yuri91:
Hi! I would like to do a brief update about my work on the branch hinting proposal in the next meeting.

### DIFF
--- a/main/2021/CG-01-05.md
+++ b/main/2021/CG-01-05.md
@@ -25,6 +25,7 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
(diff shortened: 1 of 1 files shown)


dtig:
@yuri91 If this is not a time sensitive update, would you be okay presenting at the 01/19 meeting instead? The meeting is tomorrow, and we don't have other agenda items right now.

yuri91:
Honestly I was hoping to make some progress on the proposal by gathering some feedback on the new approach. The last meeting was more than a month ago...

RossTate:
If there's a meeting tomorrow, I was about to suggest we discuss activating the new GitHub Discussions feature for the various repos.

dtig:
Sure, sounds good. The intent of moving this was to avoid convening a large group for short meetings. Will keep the agenda as-is as the last meeting was cancelled as well.

yuri91:
thanks!

PR #695, opened 2021-01-12
Add notes for 1/12/21 GC meeting

tlively:
Thanks for presenting, @RossTate! if you could add a link to your slides, that would be great.


RossTate:
Done! #696. Please double-check that I named the presentations folder correctly and used the right naming convention.

PR #726, opened 2021-03-07
Discussion item on irreducible control flow

conrad-watt:
I'm carving out a fairly large time slot for discussion, but if too many other items come along this should probably be reduced. Conversely, I'm sure the [debate](https://github.com/WebAssembly/design/issues/796) has the potential to take up all available time, so this should probably be timetabled last. To be clear, not planning to poll anything.


dtig:
Sounds good, thanks @conrad-watt.

PR #728, opened 2021-03-08
Add Branch Hinting - Poll for Phase 2 to agenda

yuri91:
I see that there are already few PRs for agenda items on the next meeting.

This should be short, but if it doesn't fit I can move it to the following meeting.


dtig:
Let's leave it in but cap it to 5 minutes - will be added to the next meeting if we run out of time.

PR #738, opened 2021-03-16
Propose follow-on discussion on irreducible control flow

conrad-watt:


RossTate:
Thanks, @conrad-watt, for the great presentation! For the follow-on discussion, the big unknown for me is how much of the performance gains achieved by optimizing compilers *for wasm* rely on semi-structured control flow (and in what manner). I don't know if you know this information, but if not I'm hoping someone can speak to that effect.

PR #744, opened 2021-04-02
Add an agenda item for adding Lin Clark as a WASI Subgroup chair.

sunfishcode:


dtig (approved):
@sunfishcode I'm guessing this would be fairly quick - would 5-10 minutes be enough time?

sunfishcode:
I agree, 5-10 minutes sounds good.

PR #759, opened 2021-04-20
Add GC 4/20 notes

tlively:


gc/2021/GC-04-20.md
```diff
@@ -0,0 +1,105 @@
...
+- Lars Hansen
+- Slava Kuzmich
+- Derek Schuff
+- Igor Sheludko
+- Fil Pizlo
+- Keith Miller
+- Jakob Kummerow
+- Zalim Bashorov
+- Ross Tate
+- Aske
```
askeksa-google, line 50: Indeed only my first name was displayed on Zoom when you took attendance. For the notes, my full name is probably more appropriate.

PR #768, opened 2021-05-03
Update sg-5-3.md

fgmccabe:


RossTate:
I had a chance to remind myself of the specifics of @lukewagner's [higher-level design](https://docs.google.com/presentation/d/1B0ib04vTEzdrtgrACwjo3Zj2LSwLSXHIKl_5rnLq-Ww) that I iterated upon here. (Sorry to have not done so in preparation for the meeting.)

tl;dr: The iteration makes async/await matches more explicit. This has the effect of making it compatible with both typed/untyped effect-handler designs as well as both dynamically/lexically scoped effect-handler designs, which offers more (current) implementation and (long-term) design flexibility for wasm-side extensions.

For more detail, this seems to be the most relevant code sample from @lukewagner's slides:
```
// wasm.wat
(module
  (import “fetch” (func $fetch (param …) (result externref)))
  (import “arrayBuffer” (func $arrayBuffer (param externref) (result externref)))
  (func (export “run”) … (call $arrayBuffer (call $fetch …)) …)
)
// script.js
let arrayBuffer = Response.prototype.arrayBuffer;
let instance =
  await WebAssembly.instantiateStreaming(fetch(‘./wasm.wasm’), {
    fetch: new WebAssembly.Function({params:[…], results:[externref], async:true}, fetch),
    arrayBuffer: new WebAssembly.Function({params:[externref], results:[externref], async:true, method:true}, arrayBuffer)
  });
let run = WebAssembly.bindAsync(instance.exports.run);
```

Observe that, when instantiating `instance`, the engine does not know that it will be used asynchronously, and that the imports each individually specify whether they are `async` or not. This seems to pose a problem should a `bindAsync`'d `run` end up calling a JS import that was *not* declared `async` versus when the non-`bindAsync`'d `run` does the same. In particular, it seems to require all wasm-to-JS calls to check if the current stacklet is a JS one or not and, if not, search (the stack?) for the appropriate JS stacklet to switch to. In the iteration I presented, I essentially combined instantiation using `async` imports with `bindAsync` of exports into one operation that makes this boundary more explicit, avoiding the need for dynamic checks. In other words, the iteration is compatible with (but does not require) a static effect system that statically ensures code only runs on an appropriate stack.

Another concern is that, if `bindAsync` itself results in a `WebAssembly.Function` of some sort, then it seems possible for a `WebAssembly.await` event to have multiple (or no) handlers on the stack, raising the question of which to suspend up to. Any answer commits us to some form of dynamically scoped effect handlers. Unfortunately, dynamically scoped effect handlers are known to have abstraction problems (and consequently composition/migration problems), see [Abstraction-Safe Effect Handlers via Tunneling by Zhang and Myers](https://cs.uwaterloo.ca/~yizhou/papers/abseff-popl2019.pdf) for more info. The iteration I presented ensures that exactly one handler is on the stack when a conceptual `await` event occurs, and consequently is forwards-compatible with both dynamically scoped effect handlers and lexically scoped effect handlers (the latter of which the paper recommends due to ensuring abstraction properties).

Hopefully that explains/motivates the changes I made. Sorry, this wasn't meant to be the focus of the presentation. I felt a valid response to such a plan would be "it really depends on the style of JS API&mdash;I'd want to know that we're not accidentally/implicitly limiting our options", so I wanted people to have a JS API strategy for the discussion that they could see/check (hopefully) does not limit options. Unfortunately illustrating that ended up taking most of the time.

lukewagner:
Thanks, it's much nicer to have explicit diffs and reasoning broken out like this.  I think there are arguments against each of the points you've made that are based on how the JS API and JS engines work in practice that you're not aware of, but a random PR in the meetings repo isn't a good place to have that technical discussion.  But hopefully you can appreciate here my meta-point that introducing a wholly different JS API without acknowledging the preceding work or any of the reasoning behind the proposed changes seemed weird.

Actually, thinking about where is the best place to discuss these JS API design questions, iiuc, there's not a feature repo for "stack switching", just a subgroup and few high-level design-repo issues.  @fgmccabe do you think we should create one?

fgmccabe:
Actually, there is now a repo for stack switching: https://github.com/WebAssembly/stack-switching

lukewagner:
Oh whoops, I missed that, thank you.

PR #773, opened 2021-05-07
Move WASI meeting minutes to meetings repo

linclark:
The WASI SG is moving all meeting agendas and notes to this repo to align with the process that has emerged since the SG was created.

I don't think there's any way to keep the commit history here, but if any git wizard knows of a magical way to do that (that isn't too onerous), I'm open to trying it out.


sunfishcode (approved):
Looks good! I expect the commit histories for these aren't very interesting, and if we ever do need them, they'll still be in the WASI repo, so I think this looks good.

PR #775, opened 2021-05-10
Add agenda item relevant for second CM poll 05-25

dcodeIO:
I have pre-recorded a presentation on why I think that we should not proceed with the next steps as presented in "Scoping and Layering the Module Linking and Interface Types proposals", that I would like to play at a meeting for the necessary context. I am aware that my presentation does not fit into this meeting's time slot anymore, but as it is highly relevant for the poll, it needs to be scheduled before somehow. Please feel free to re-schedule as needed :)

### DIFF
--- a/main/2021/CG-05-25.md
+++ b/main/2021/CG-05-25.md
@@ -25,6 +25,7 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
+    1. Present [WebAssembly, UTF-8 and Web Platform APIs - Or: Please don't break the Web 🥺](https://youtu.be/IhvwO0zkbXU) (Daniel Wirtz) [15 mins]
     1. Follow-up discussion from the [CG-04-27 presentation](https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8) and [design/#1415](https://github.com/WebAssembly/design/issues/1415) (20 min)
        1. Poll: Does the proposed direction sound good?
        1. Poll: Should we proceed with the [next steps](https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8/edit#slide=id.gcd82cd8251_3_6)?
--- a/main/2021/CG-05-25.md
+++ b/main/2021/CG-05-25.md
@@ -25,7 +25,7 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
-    1. Present [WebAssembly, UTF-8 and Web Platform APIs - Or: Please don't break the Web 🥺](https://youtu.be/IhvwO0zkbXU) (Daniel Wirtz) [15 mins]
+    1. Present [WebAssembly, UTF-8 and Web Platform APIs](https://youtu.be/IhvwO0zkbXU) (Daniel Wirtz) [15 mins]
     1. Follow-up discussion from the [CG-04-27 presentation](https://docs.google.com/presentation/d/1PSC3Q5oFsJEaYyV5lNJvVgh-SNxhySWUqZ6puyojMi8) and [design/#1415](https://github.com/WebAssembly/design/issues/1415) (20 min)
        1. Poll: Does the proposed direction sound good?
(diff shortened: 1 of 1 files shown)


main/2021/CG-05-25.md
```diff
@@ -25,6 +25,7 @@ Installation is required, see the calendar invite.
 1. Adoption of the agenda
 1. Proposals and discussions
     1. Review of action items from prior meeting.
+    1. Present [WebAssembly, UTF-8 and Web Platform APIs - Or: Please don't break the Web 🥺](https://youtu.be/IhvwO0zkbXU) (Daniel Wirtz) [15 mins]
```
Ms2ger, line 28: The video appears to be "private"

dcodeIO: I haven't made it public yet. I don't see a reason why I couldn't make it public ahead of time, so everyone can already watch when they have time etc. Would know about scheduling first, though, in case I have to edit.

conrad-watt, line 28: I would like this title to be changed. None of the proposals/features brought up [here](https://github.com/WebAssembly/design/issues/1407) "break the Web" - this is an absolute principle we must not violate in our Web-facing semantics. Characterising the discussion in this way is not a productive first step.

I agree that there is legitimate debate to be had about performance and ecosystem fragmentation, which I would welcome so long as we can avoid hyperbole.

dcodeIO: In the presentation, I mean and hope to manage to express it as a friendly wink lessening the tension from foregoing heated discussions (I read the "pleading face" for instance), while at the same time hinting at the problem the presentation will discuss, as it is indeed pointing out where the Web's design principle of backwards compatibility may be violated.

But thank you for the heads up! I will remove the subtitle from this PR however, as I agree that, without context, it is easily interpreted this way.

dtig:
The full hour for this meeting is accounted for, so we may not be able to get to this item on 5/25. The one thing I could do is to reorder the agenda, so if the Wasm Signatures discussion doesn't take the full 40 minutes, we could squeeze in some discussion time, though I suspect that we may need more time here regardless. 

We haven't had pre-recorder presentations before, could you clarify whether the 15 minutes is the length of the presentation, or presentation + expected discussion time?

dtig:
After offline follow up, I've reset the state on https://github.com/WebAssembly/design/issues/1407 to the last productive point, and unlocked the issue. Please feel free to continue the discussion there. 

From a procedural perspective, I'd recommend grabbing a slot at the meeting following this one. Depending on the discussion at the meeting, if we find that we can't reach consensus on the direction, we may push the poll, or continue the discussion at the next meeting (we've done this quite often in the past).

As we haven't had pre-recorded content before, I'm wondering if it will be worthwhile to open the link so folks can watch it before the 5/25 meeting, and have a slot for discussion at the 6/8 meeting that can be led by you.

dcodeIO:
Thank you @dtig, appreciate it! Since this has created some interest, I have been working on improvements to the presentation given feedback from peers who have already watched it or I otherwise talked to. It will get a little longer (and I hope much more friendly), so moving to the next meeting seems good :) If the vote happens earlier, I may then vote with "strongly against", with formal objection, but not because I'd be seeking chaos, but rather because I would like to give everyone the opportunity to make an informed decision. Would that be OK, or be considered bad practice?

penzn:
Just my personal opinion - pre-watching a pre-recorded video is more productive than using live meeting time to do it. Given the activity around this, I doubt that many (if anybody at all) will come to the discussion unprepared.

dtig:
> Thank you @dtig, appreciate it! Since this has created some interest, I have been working on improvements to the presentation given feedback from peers who have already watched it or I otherwise talked to. It will get a little longer (and I hope much more friendly), so moving to the next meeting seems good :) If the vote happens earlier, I may then vote with "strongly against", with formal objection, but not because I'd be seeking chaos, but rather because I would like to give everyone the opportunity to make an informed decision. Would that be OK, or be considered bad practice?

That would be okay, and I'd like to add that collecting dissenting statements is also the intent of the polling process, so voting against with concrete reasons is an example of how the process should be working.

dcodeIO:
@penzn I would agree, yeah, but would also kinda like to give the presentation at a meeting, be it only for the record that it has been held. Do you know if there are processes in place that would allow the "for the record" bit otherwise?

conrad-watt:
@dcodeIO in this scenario I imagine we'd still schedule discussion time in a meeting. Making the talk available ahead of time would allow the timeslot to be 100% follow-on discussion; the meeting notes could still link to the talk.

dtig:
> @dcodeIO in this scenario I imagine we'd still schedule discussion time in a meeting. Making the talk available ahead of time would allow the timeslot to be 100% follow-on discussion; the meeting notes could still link to the talk.

+1 to this, linked as it is in this PR in the agenda, possibly with a note for context, and added again to the notes will address the "for the record bit".

PR #795, opened 2021-06-04
Create discussions.md for constructive CG discussions

dtig:
Adding a document with suggestions for constructive CG discussions. This is not specific to meetings, but we don't have a good place to put general process documents, so this seems like a good start.

Thanks @tlively for collaborating on this document.


penzn (approved):

tlively (approved):
🎉

dschuff (approved):

PR #838, opened 2021-07-26
Add an agenda item for updates on interface types

alexcrichton:
It was suggested to me that now's probably a good time to talk about
developments in the interface types proposal and how it can affect WASI,
and I agree it's probably good to talk about as well!


linclark:
This sounds great, thanks!

PR #841, opened 2021-07-29
Add vote for wasi-http to WASI 2021-08-12 agenda.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

### DIFF
--- a/wasi/2021/WASI-08-12.md
+++ b/wasi/2021/WASI-08-12.md
@@ -28,4 +28,6 @@ The meeting will be on a zoom.us video conference.
 1. Announcements
     1. _Sumbit a PR to add your announcement here_
 1. Proposals and discussions
-    1. _Sumbit a PR to add your agenda item here_
+    1. New proposal: wasi-http (split from Proxy-Wasm). (10 min)
+        1. Vote: Approve for Phase 1?
+    1. _Sumbit a PR to add your announcement here_


wasi/2021/WASI-08-12.md
```diff
@@ -28,4 +28,6 @@ The meeting will be on a zoom.us video conference.
 1. Announcements
     1. _Sumbit a PR to add your announcement here_
 1. Proposals and discussions
-    1. _Sumbit a PR to add your agenda item here_
+    1. New proposal: wasi-http (split from Proxy-Wasm). (10 min)
+        1. Vote: Approve for Phase 1?
```
PiotrSikora, line 32: @linclark technically, Proxy-Wasm as a whole was already approved in 2020-04-09 meeting: https://github.com/PiotrSikora/meetings/blob/main/wasi/2020/WASI-04-09.md#meeting-notes

linclark: Ah, good point! That was before my time as co-chair, so I didn't realize that vote had already happened.

Given that, I feel like we can create the repo. Would you still like to give an update in this meeting?

PiotrSikora: Let's skip it until there is a meaningful update to talk about.

PR #880, opened 2021-09-21
Add an item for simple proposals to 2021-09-28

penzn:
For #857


dtig (changes requested):
Thanks @penzn! Could you also time box this discussion?

penzn:
Done.

dtig (approved):

PR #888, opened 2021-09-28
Slides for Flexible Vectors update

penzn:


penzn:
This is probably still tentative, since it is the last topic on the agenda ;)

PR #959, opened 2022-02-02
Discuss: It's hard to contribute on proposals.

cardil:
I find it difficult to contribute to existing proposals. They are a forked WASI repos with additional files.

There's no space for discussion on them, other than creating a PR that targets the fork. I'd like to have a space to discuss first (Github discussions, maybe?!?). Even, opening an PR is hard as it may interfere with the main maintainer commits.

I think the better might be to commit (by PR) a stub of a proposal first. So, then multiple people could provide parts of the proposal in smaller chunks.

### DIFF
--- a/wasi/2022/WASI-02-10.md
+++ b/wasi/2022/WASI-02-10.md
@@ -28,4 +28,5 @@ The meeting will be on a zoom.us video conference.
     1. _Sumbit a PR to add your announcement here_
 1. Proposals and discussions
     1. Discuss network APIs split. [PR](https://github.com/WebAssembly/WASI/pull/461) - Dave Bakker (@badeend)
-    1. _Sumbit a PR to add your agenda item here_
+    2. It's hard to contribute on proposals. [PR](https://github.com/WebAssembly/WASI/pull/xxx) - Chris Suszynski (@cardil)
+    3. _Sumbit a PR to add your agenda item here_


sbc100:
I think the idea (as with WebAssembly proposals) is that one can open issues and PR in the forked repository, keeping the discussion local to the proposal.   We tend to use github issues for "discussions" but maybe we can move to the official github discussions thing instead?  I'm not sure it make that much difference.  Happy to discuss further in the meeting of course.

linclark:
Heads up, this process is on the verge of changing already.

This has been discussed in multiple WASI meetings, and in the issue queue with https://github.com/WebAssembly/WASI/pull/451 (which I am planning to merge today).

sbc100:
> Not all forks have issues enabled. I'm interested the most in https://github.com/badeend/WASI-Networking, but I saw a couple of other ones, do not have issues enabled as well.

That sounds like a bug with those proposals.. for sure proposal should have issues enabled.

linclark:
Totally agreed—all proposals in the Proposals list should have issues enabled. Thanks for raising that. I'll go through that list and contact the champions wherever that's not the case right now.

I merged https://github.com/WebAssembly/WASI/pull/451, and with that, we won't be using forks anymore but instead will use this [proposal template](https://github.com/WebAssembly/wasi-proposal-template).

@cardil Why don't you take a look at the updated WASI repo and the proposal template and see whether you still think there are still issues to be discussed in the meeting. If you do still think it needs a synchronous discussion after that, just ping here before next Wednesday and I'll merge this PR.

cardil:
I'll take a look. Converting to draft for now.

PR #965, opened 2022-02-08
Create GC-02-08.md

tlively:


gc/2022/GC-02-08.md
```diff
@@ -0,0 +1,266 @@
...
+
+BT: Regardless of whether it's a dimension or a possible feature of unions, you'd have to pull it out of the hierarchy.
+
+ASC: That would be necessary to avoid checking for i31 when downcasting from eq or any.
+
+AR: If you want to avoid those checks you can just avoid using `any` and `eq` (those are just unions).
+
+TL: AR, you said that you couldn't use pointer tagging with these unions? Doesn't that make i31 useless?
+
+CW: Downcasting would perform the check and could change the representation.
```
conrad-watt, line 194:
```suggestion
CW: Pointer tagging should still work? Downcasting won't change representation, values of precise types will keep their tags.
```

This better captures what I said during the meeting, although I don't think this point was discussed further so I could have said something incorrect.

tlively: Great, thanks for the edit!

PR #992, opened 2022-03-09
Some corrections/refinements on GC 02/22 meeting notes

rossberg:
Just noticed that I had created this patch last week but forgot to create an actual PR.

### DIFF
--- a/gc/2022/GC-02-22.md
+++ b/gc/2022/GC-02-22.md
@@ -59,11 +59,11 @@ TL: Binaryen & V8 have been experimenting with versions of the instructions that
 
 TL: There seem to be two arguments for RTTs: JS API and generics. For the JS API, I'd suggest we think about that without RTTs being a hard requirement.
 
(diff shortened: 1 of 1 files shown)


gc/2022/GC-02-22.md
```diff
@@ -59,11 +59,11 @@ TL: Binaryen & V8 have been experimenting with versions of the instructions that
...
 
 BT: Cost argument is a sliding scale. If we applied it universally, e.g. call_indirect would be split into multiple operations. My reason for wanting to keep RTTs is that if they are implicit, that might be an extra load from the instance object. But that depends on implementation details, so the explicit cost argument cuts both ways.
 
-AR: Fair, it is a blurry line. My argument is that, if it's something we'll need anyway, we should add it now since it also fits with the cost model. call_indirect is an interesting example, I don't think anyone was really happy with how that came out; if we had to do it again I could imagine it being designed differently.
+AR: Fair, it is a blurry line. My argument is that, if it's something we'll need anyway, we should add it now since it also fits with the cost model. call_indirect is an interesting example, I don't think anyone was really happy with how that came out; if we had to do introduce it at this point I could imagine it being designed differently, e.g., in terms of casts.
```
tlively, line 66:
```suggestion
AR: Fair, it is a blurry line. My argument is that, if it's something we'll need anyway, we should add it now since it also fits with the cost model. call_indirect is an interesting example, I don't think anyone was really happy with how that came out; if we had to introduce it at this point I could imagine it being designed differently, e.g., in terms of casts.
```

tlively (approved):

PR #1005, opened 2022-04-05
Add a WASI URL interface agenda item

kulakowski-wasm:


kulakowski-wasm:
@linclark I believe this is your meeting?

linclark:
Indeed it is! Unfortunately we can't do a poll unless we've provided 7 days notice, but we can talk about it in this meeting, and then conduct a poll in the next meeting.

PR #1012, opened 2022-04-13
GC agenda items for `nullref` and cast refactoring

tlively:


rossberg:
@tlively, thanks, makes sense to discuss this. However, I'm OOO the week of Easter, so won't be able to attend the next meeting. Do you think we can push the discussion to the one after?

tlively:
@rossberg, thanks for the heads up! Yes, let's push this discussion, then.

PR #1014, opened 2022-04-15
Add stringrefs to GC subgroup agenda, 19 April 2022

wingo:
The [stringrefs proposal](https://github.com/wingo/stringrefs/) defines a facility for representing strings.  I plan on proposing it for phase 0 in the CG meeting on 26 April.  Though it doesn't explicitly depend on the GC proposal, it is in the same family, so I thought a brief intro to the GC group might be useful.


tlively (approved):
LGTM!

PR #1015, opened 2022-04-15
Add stringrefs to CG agenda, 26 April 2022

wingo:
The [stringrefs proposal](https://github.com/wingo/stringrefs) defines a facility for representing reference-typed strings.  I've been gathering early feedback and working on it off and on for some months now.  I would appreciate any feedback from the CG.

As regards polls, I had planned on going to phase 0, but I see that is effectively already done.  Depending on the interest and feedback, it might be appropriate to poll for phase 1 already.


dschuff (approved):
Yeah, I think "possible poll for phase 1"  is fine to have here.

PR #1018, opened 2022-04-21
Add WASI 04-21 meeting notes

linclark:


wasi/2022/WASI-04-21.md
```diff
@@ -31,3 +31,64 @@ The meeting will be on a zoom.us video conference.
...
+
+One example where JS URL isn’t good general purpose design… there’s an accessor for part. That returns a string instead of a number. JS doesn’t have an option type, so if it isn’t set then would be null, which would coerce to 0. We’d probably go with option, and then to implement in JS you’d need a tiny bit of glue code.
+
+**George Kulakowski:** Yes, I do think I mentioned this would err on the side of typed. Someone was asking about why not more closely hewed to browser. Using strings wouldn’t help because you still have to go to your native implementation.
+
+**Lin Clark:** Any more questions?
+
+Ok, consensus poll. Anyone object? [long awkward silence]
+
+Great, we have a new API!
```
trusktr, line 86: Why not align with the existing `URL` spec rather than make a new spec?

https://developer.mozilla.org/en-US/docs/Web/API/URL

trusktr: The idea should be that binding for one language work the same regardless if using `URL` in a browser, or `URL` in WASI. It seems like a good idea for developer consistency.

trusktr: I see discussion regarding this above, but still. 

Or is the plan that WASI will automatically map to URL behind the scenes, and people will only write for WASI and it works everywhere?

PR #1024, opened 2022-05-04
Update WASI-05-05.md

AlexEne:
Add a proposal to kickstart a discussion over a native threads API.

This is my first time doing this, I've read the docs, and started a (very WIP) repository here: https://github.com/AlexEne/wasi-native-threads/blob/main/README.md

I hope I haven't missed any step.
I will prepare a high-level presentation for the meeting. 15 min should be ok?


linclark:
Perfect! We won't be able to poll tomorrow, but you can definitely kick off the discussion, and then we can poll in the next meeting. And 15 mins is great. We have a light agenda, so you can also take more time if needed.

PR #1029, opened 2022-05-10
Update CG-05-10.md

titzer:
Add item for discussion WebAssembly Research.


dschuff (approved):
I guess this is a bit last minute, but it seems likely we'll have time.

PR #1035, opened 2022-05-16
Add discussion item for gc typesystem mode

titzer:


tlively (approved):
Great, thanks!

tlively (approved):
Great, thanks! I added timing notes (30 minutes for each discussion) as well, although I expect the discussions to bleed into one another.

PR #1036, opened 2022-05-18
Add documentation on how to start a proposal

tlively:
The official phases document lays out the basic requirements for starting a proposal, but it will be useful to have more detailed and opinionated documentation as well.


dtig (approved):
Thanks for putting this together!

PR #1058, opened 2022-06-15
discussion item around WIT in WASI

ricochet:
We have been treating WIT as an external tool that supports working with the component model. In a lot of ways, however, WIT is an IDL that may benefit from standardization. The closest body for this is this group.


linclark:
Ah, good call! We had originally intended for it to be standardized in the component model, but I just checked with @lukewagner and he says that with the benefit of time passing, he actually thinks this is the right place. We can discuss more in the meeting.

PR #1060, opened 2022-06-21
Fix June 21 meeting link

takikawa:
There's currently a broken June 14 link on the main page, but I believe it was meant to point to the June 21st meeting for tomorrow.


dtig (approved):
Thanks!

PR #1067, opened 2022-06-30
Update 6-27-22.md

fgmccabe:
Please add notes if you took any. This meeting was not recorded.

### DIFF
--- a/stack/2022/6-27-22.md
+++ b/stack/2022/6-27-22.md
@@ -8,6 +8,19 @@
 
 
 ## Participants
+Francis McCabe
+Daniel Hillerström
+Ross Tate
+Rick Battagline
+Luke Wagner
+Maciej Pirog
+Ilya Rezvov
+Thibaud Michaud
+Ryan Hunt
+Deepti Gandluri
+Paul Schoenfelder
+Petr Penzin
+
 
 ## Agenda items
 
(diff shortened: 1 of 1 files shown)


stack/2022/6-27-22.md
```diff
@@ -8,13 +8,29 @@
...
 
 ## Agenda items
 
 1. Continued discussion on task-based stack switching. (AMA)
 
 ## Meeting Notes
 
+Fgm gave a rerun of previous presentation on task based stack switching.
+There was some discussion about how erlang would integrate. PS expressed a concern that, since there is no stack walking in this proposal, Erlang implementations would still need to use a shadow stack to enable support for GC.
+PP expressed some concerns about stack management as it relates to having many threads.
```
penzn, line 33: I was asking how they manage shadow stack for many threads, and the answer was something along the lines of making and estimate how much each would need and then allocating shadow stack entries for them, as opposed to classic Erlang when it attempts to allocate absolute minimum and then grow the areas.

I also made a comment at the end that exact style of shadow stack support is orthogonal to this proposal.

I am also in favor of this proposal as it seems simple enough and takes important details into account (I said something along those lines, but can't remember what anymore).

penzn (commented):
Sorry, this is from memory, as I also didn't take notes. Maybe I can help take notes sometimes.

PR #1074, opened 2022-07-17
Add update on internval primitives

KloudKoder:
In the year since last discussion, there have been a few developments which should enable the CG to make a final decision on whether or not to develop this functionality. I'd like to share those news items and invite debate.


conrad-watt:
Currently this PR modifies a WG agenda - I think it should instead be scheduling something as a CG agenda item?

PR #1083, opened 2022-08-04
Agenda item to announce Conrad Watt as new co-chair

dtig:


dschuff (approved):

conrad-watt:
I'm very grateful to be financially supported in this role by Microsoft and Mozilla. I'll also acknowledge this arrangement during the meeting.

PR #1094, opened 2022-08-16
Added item to the Agenda

syrusakbary:
TLDR: this probably wouldn't have needed to be an item on the agenda.

Just some responsiveness on the [original issue](https://github.com/WebAssembly/WASI/issues/492) would have been enough. However, I believe is important to reach an "official" agreement on what the expected behavior is, and we should at least cover it on the agenda as means to have the question answered

### DIFF
--- a/wasi/2022/WASI-08-25.md
+++ b/wasi/2022/WASI-08-25.md
@@ -27,4 +27,12 @@ The meeting will be on a zoom.us video conference.
 1. Announcements
     1. _Sumbit a PR to add your announcement here_
 1. Proposals and discussions
-    1. _Sumbit a PR to add your discussion item here_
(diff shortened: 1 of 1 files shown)


linclark:
@syrusakbary TBH, I'm somewhat confused by the stance you're taking here. Your comment in the original issue was the first comment, and it called for it to be added to the agenda. If you felt it was better to address it async in the issue, then you should have asked for more discussion in the issue rather than proposing it as an agenda item.

This message expresses impatience with the response time on the related issue. That issue was filed in early August. I would suggest that you show a little more grace to the contributors—being August, it is prime PTO time for many of the individuals who contribute to the WASI subgroup.

As such, I will ask for this to be addressed in the issue instead of as an agenda item.

PR #1097, opened 2022-08-25
Add notes for 8/23 GC meeting

dtig:


dtig:
I may have missed some of the points that @rossberg and @titzer made towards the end, please add/edit as needed. Thanks!

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+AZ: Can you clarify why it’s easier to add unannotated versions later vs. adding the annotated versions later?
+
+AR: When we run into problems with lack of annotations, we can’t add them because of backwards compatibility, because we’re tied into supporting these instructions forever.  If there’s interference with some feature we want to add, then we basically a bad place. We have to resolve this somehow. The other way is more robust, the size optimization is always something you can add at later.
+AZ: Why do we think we’d be at a point of time in the future where we think the risk is lower vs. now? Do you think we’ll come to a point where we’ve implemented all the tricky things in the type system? That seems unlikely. We’d always have the argument against adding them, we’ll always thinking it’s risky to add. Worry is always that we will keep suffering the 5% loss indefinitely, cost can be significant in the long run, and we’re choosing to opt into it, we might always say it’s risky to add them
```
tlively, line 89:
```suggestion

AZ: Why do we think we’d be at a point of time in the future where we think the risk is lower vs. now? Do you think we’ll come to a point where we’ve implemented all the tricky things in the type system? That seems unlikely. We’d always have the argument against adding them, we’ll always thinking it’s risky to add. Worry is always that we will keep suffering the 5% loss indefinitely, cost can be significant in the long run, and we’re choosing to opt into it, we might always say it’s risky to add them
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+BT: We are designing a format that has many different purposes, it’s a code format for interchange, we should prioritize all the tools that process that format, producers and consumers, and I don’t see how we justify that.. We could make the reverse argument, why aren’t there annotations that would be useful for tool authors and engines? From a different point of view that might be a hostile design decision
+
+AR: This is very real feedback from people that work in the research space, one of the top things that makes their life easier with Wasm was that there is no overload, which is the trivial case of the problem we’re discussing right now. We’re talking about a much more inconvenient relaxation.. I would also like to push back on the size arguments, it falls into the category of premature optimization, we haven’t really optimized for size as much previously, putting in adhoc solutions to individual small things, we could have had a const 0 inaructions, but we decided against it, we don’t have a negate instructions, and all these things we never decided to size optimized, we should do a more targeted approach instead of adhoc ideas, we should stay consistent with that
```
tlively, line 97:
```suggestion
AR: This is very real feedback from people that work in the research space, one of the top things that makes their life easier with Wasm was that there is no overload, which is the trivial case of the problem we’re discussing right now. We’re talking about a much more inconvenient relaxation. I would also like to push back on the size arguments, it falls into the category of premature optimization, we haven’t really optimized for size as much previously, putting in ad hoc solutions to individual small things, we could have had a const 0 instruction, but we decided against it, we don’t have a negate instruction, and all these things we never decided to size optimize, we should do a more targeted approach instead of ad hoc ideas, we should stay consistent with that.
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+AZ: Fair point that the compression could be done at another level, we haven’t focused on micro optimizing things. Want to push back against some of these, while in theory we could do something in the space, but it would take time, it’s unclear that we would actually do this. Unclear that we would actually have the resources to work on this, there are things we could have done the past, but did not but those are things  (const 0, const 1) that compress well, on the other hand, type annotations are high entropy thing and have a higher cost, this stuff does matter, we have a lot of use cases on teh web where there are limited on the size of the download, while it doesn’t matter in some countries with unlimited bandwidth, other countries it isn’t. 5% is significant, lots of efforts to reduce code size overall and ever % matters. IT is likely that if we don’t save this now, it’s possible that we’ll be stuck with this for a very long time and that would be wasteful.
```
tlively, line 101:
```suggestion
AZ: Fair point that the compression could be done at another level, we haven’t focused on micro optimizing things. Want to push back against some of these, while in theory we could do something in the space, but it would take time, it’s unclear that we would actually do this. Unclear that we would actually have the resources to work on this, there are things we could have done the past, but did not but those are things  (const 0, const 1) that compress well, on the other hand, type annotations are high entropy thing and have a higher cost, this stuff does matter, we have a lot of use cases on the web where there are limited on the size of the download, while it doesn’t matter in some countries with unlimited bandwidth, other countries it isn’t. 5% is significant, lots of efforts to reduce code size overall and ever % matters. It is likely that if we don’t save this now, it’s possible that we’ll be stuck with this for a very long time and that would be wasteful.
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+DG: We’ve had discussions on both sides of this, one of the things that I’d like to circle back on, was that there are future features  that there is a point of time in the future that we might feel comfortable about not having type annotations, you mentioned type parameters, maybe we can dig into that a little more what are other examples of future features? It would be nice to quantify that, at what point of time do we think the risk is lower, or is there a feature set you have in mind?
```
tlively, line 103:
```suggestion
DG: We’ve had discussions on both sides of this, one of the things that I’d like to circle back on, was that there are future features that there is a point of time in the future that we might feel comfortable about not having type annotations, you mentioned type parameters, maybe we can dig into that a little more what are other examples of future features? It would be nice to quantify that, at what point of time do we think the risk is lower, or is there a feature set you have in mind?
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+AR: It’s really a spectrum, certainty is increasing monotonically, at some point we can have a more informed decision about this. What the features would be, or what the timeline would be, I would make no predictions at this point given how long we’ve worked on at this point
+
+DG: phrased this wrong, maybe looking for more of an opinion, the point that this could be unbounded w.r.t to time/features it seems uncomfortable at a time that we see the binary size being bloated that we want to make sure that we’re not adding size increase in unless necessary
+
+FM: What would be the rationale for having type parameters? I’m not sure that there is one
+
+AR: Currently there is no way to create code that operates the same on any reference, where you don’t really care  what the reference is. This is a frequent thing for compilers, in Wasm you can’t express them because every reference has to be concretely defined, i view  this feature as giving a way to generate this kind of code that native code compilers are able to generate
```
tlively, line 111:
```suggestion
AR: Currently there is no way to create code that operates the same on any reference, where you don’t really care  what the reference is. This is a frequent thing for compilers, in Wasm you can’t express them because every reference has to be concretely defined, I view this feature as giving a way to generate this kind of code that native code compilers are able to generate
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+BT: I’ll +1 to that, the tools want annotations and we might want to keep the annotations indefinitely, but have the tools do the compression. 
+
+DG: That’s a reasonable answer, that assumes that we do the work to handle compression at a different layer, and that the tools do the right thing and we have the tools in place to do that right?
+
+BT: Right, Isee layer 1 compression as the thing that’ll give us the best of both worlds. We could use the Wasm type system as the type compression, and still have type annotation in the bytecode for future tools, and future feature extensions
```
tlively, line 123:
```suggestion
BT: Right, I see layer 1 compression as the thing that’ll give us the best of both worlds. We could use the Wasm type system as the type compression, and still have type annotation in the bytecode for future tools, and future feature extensions
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+DG: Any other thoughts? 
+
+CW: Restate something came up implicitly, in a couple of years from now, if we do have a more solidified view of what extensions we want in the future, we can still add annotationless versions in the future with essentially the same code size savings. 
+
+AR: That’s the conservativeness argument
+
+DG: It doesn’t have to be an example for this particular set, but is there an example that you could point to of something that was broken, looking for something that’s not as hypothetical as future problems, extensions, but even a toy type system example that can run into this problem
+
+AR: Select instruction is an example it didn’t have annotations, and once we added subtyping we realized we have to compute LUBs now we did this hack, where it was moderately ugly, we’ve already run into this problem, the br_table was also simple but hack not so ugly, simple example that everyone can understand. With the kind of things we’re talking about now, the problems would be much more intricate. IT’s not completely generic types, it’s very specific types
```
tlively, line 133:
```suggestion
AR: Select instruction is an example it didn’t have annotations, and once we added subtyping we realized we have to compute LUBs now we did this hack, where it was moderately ugly, we’ve already run into this problem, the br_table was also simple but hack not so ugly, simple example that everyone can understand. With the kind of things we’re talking about now, the problems would be much more intricate. It’s not completely generic types, it’s very specific types
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+
+BT: Ran into this problem for func.bind, which has a synthesized type, we run into the same issue with any type of inference that it creates types that you haven’t written down. It’s because we have this subtyping relationship that we have to declare, if you haven’t written one of the subtypes in the chain, it’s essentially a synthesized type, and how is that type canonicalized? We don’t have to mandate type canonicalization
+
+AR: LUBs only work when you have proper lattice in the type system, or at least a semi lattice. For br_table you need GLBs (?) so you would need a full lattice, that’s a severe constraint. There are cases where you can’t synthesize a proper type, there is no principle type that you could synthesize, it’s not clear how you would solve this problem
```
tlively, line 141: @rossberg, why would br_table need GLBs?

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,187 @@ Installation is required, see the calendar invite.
...
+FM: We looked at doing macros, the macro question keeps coming up, but as it happened that I was working on CSP at the time, it causes sufficient pain to drop it at the time, it’s possible we could fix it
+
+CW: To me the core of what’s going on is the meta question, of when we want to do that worrying, can we defer the eat the 5% and come back to it later
+
+BT: Is 5% shippable? 
+
+AZ: 5% is not going to make the difference between shipping and not, it could be something we suffer for a long time, it affects everyone, not just one subset of people working on toolchains/specs, there are risks to people across the board and we have to consider that
+
+BT: I do want to discount that, it is a binary question of can we ship, that determines whether we want to remove annotations
+AZ: I think we could ship either way, just one way would be less good
```
tlively, line 198:
```suggestion

AZ: I think we could ship either way, just one way would be less good
```

tlively (commented):
Thanks for running this meeting and taking notes, @dtig!

gc/2022/GC-08-23.md
rossberg: @tlively, this was a problem we ran into when we introduced reference types and that we could only fix by adding bottom, see my presentation from June 2019: https://github.com/WebAssembly/meetings/blob/main/main/2019/presentations/2019-06-rossberg-reference-types.pdf

FWIW, people keep being surprised by this one, see e.g. [this](https://github.com/WebAssembly/spec/pull/1521#issuecomment-1220548348) very recent comment.

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,189 @@ Installation is required, see the calendar invite.
...
+
+JK: To validate the annotation, which you have to do in an engine, you do have to check the type on the stack, for example, you have to check is there even a struct to do struct.get on that, and if you can tell easily enough whether you have a struct at all, then presumably you can always tell whether that has a type parameter or not 
+
+AR: It’s more complicated than that. In general, do you have a principled type or not? Is there a most generic type that covers all possibilities. If you don’t have that then you need back checking in general. This is a thing you don’t want to do. To even decide whether your type is principled, that’s not something you can do without backtracking either. You have to explore all the search space where you could have decided otherwise. I have seen systems that track the principality of the types used, but that is very research,  I don’t know how well it generalizes. In general, the assumption that you can make this distinction doesn’t work
```
rossberg, line 154:
```suggestion
AR: It’s more complicated than that. In general, do you have a principal type or not? Is there a most generic type that covers all possibilities. If you don’t have that then you need back tracking in general. This is a thing you don’t want to do. To even decide whether your type is principal, that’s not something you can generally do without backtracking either. You have to explore all the search space where you could have decided otherwise. I have seen systems that track the principality of the types used, but that is very researchy,  I don’t know how well it generalizes. In general, the assumption that you can make this distinction doesn’t work
```

gc/2022/GC-08-23.md
```diff
@@ -33,3 +33,189 @@ Installation is required, see the calendar invite.
...
+
+BT: In practise what would happen is that the annotation less instruction become less and less useful over time, because you add features that can’t use them, and then you have to come up with rules for when  they can be used and not and they can be very subtle
+
+AR: Because you’re very likely to use properties like substitutibility, if you have very adhoc restrictionson what types are allowed somewhere, one essential rule of subtyping is that whenever you can use a type, you can always use a subtype, if you break the property, you’ve broken the entire type system, so thes  rules have to work in a way that you don’t break the property, I’m not sure that it’s possible
```
rossberg, line 158:
```suggestion
AR: Because you’re very likely to lose properties like substitutability, if you have very adhoc restrictionson what types are allowed somewhere. One essential principle of subtyping is that whenever you can use a type, you can always use a subtype. If you break the property, you’ve broken the entire type system, so these rules have to work in a way that you don’t break the property, I’m not sure that is possible in general
```

gc/2022/GC-08-23.md
tlively: Thanks, that makes sense.

PR #1099, opened 2022-08-26
Add poll for Spiderlightning to agenda

kulakowski-wasm:
@linclark @Mossaka I believe we decided to do this for the next meeting? If so, I wanted to make sure it's on the agenda with sufficient notice. Lin, if so, is this correctly phrased? Does it need to be more specific about e.g. which phase, or link to any materials?


danbugs:
That looks good to me, @kulakowski-wasm! Please, let us know if there any actions required from the SpiderLightning side.

linclark:
Thanks for adding this, @kulakowski-wasm! 

It would be great to link to the slides, and we should note that it is a poll to move to Phase 1.

kulakowski-wasm:
Done!

PR #1104, opened 2022-09-08
Update CG-09-27.md

fgmccabe:
Added entry for update on Stack Switching


dtig:
I realized that the formatting of this agenda doc was off, fixed and updated time slot to 1 hr instead of long.

PR #1112, opened 2022-09-21
Add some agenda items

rossberg:


conrad-watt:
Are @fgmccabe, @slindley and @dhil ok with these scheduling restrictions (I assume this is to facilitate remote participation from Edinburgh)?

For reference (and @fgmccabe please chime in here), are fibers and typed stack switching in some way mutually exclusive or could we imagine both being ultimately standardised?

rossberg:
@conrad-watt, yes, @dhil will be on site, but @slindley can't come. I took the liberty to assume that @fgmccabe won't mind either way, but let us know if that's a problem.

The proposals are either mutually exclusive or we find a way to converge them. Personally, I doubt we'll be in a suitable state for voting.

tlively:
GC phase 3 poll at the meeting SGTM!

fgmccabe:
yes

PR #1113, opened 2022-09-21
Add items for GC Roots and for wasm-call Performance Analysis

RossTate:
Sorry to ask to take up the whole meeting. I only just finished a major deadline, so this is the only chance I have to present ahead of the in-person meeting.


dtig (approved):

dtig:
Merging this, as the next two meetings are fully scheduled, I'll also leave a note in the scheduling section for this meeting to email the chairs in case of time sensitive items that need attention.

PR #1133, opened 2022-10-14
Add meeting notes for SIMD call 2022-10-14

penzn:


simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+AK: yes, there are classes of algos, where you can use either single or double, and others when you need one or the other. We don’t want people replacing qfma with fma unless they understand the cost.
+
+CW: wrt to your recent post, for programs that want to take mul ad to replace with qfma, do they rely error state within tolerance, or result is exactly repeatable when doing calculation twice?
+
+AK: it depends, some cases where it doesn't matter, some cases where the paths in the program differs. It is hard to know. Large class of C programs, GCC -O3 can occasionally lower a*b+c to fma. Large class of programs compiled from C to assembly with this setting, and it works. If we have set non-det for qfma, it is not end of the world. Not sure if this is a better world than list-non-determinism. What are we gaining from set non-det, while a lot of implementations will be list non-determinsm, we are losing some valuable properties, and it is worth it? Every time i need to do a feature gate, it's a major pain, as an application developer on top of Wasm, won't use any of this tricks, maybe not worth it, gets complicated fast.
```
ngzhian, line 60:
```suggestion
AK: it depends, some cases where it doesn't matter, some cases where the paths in the program differs. It is hard to know. Large class of C programs, GCC -O3 can occasionally lower a*b+c to fma. Large class of programs compiled from C to assembly with this setting, and it works. If we have set non-det for qfma, it is not end of the world. Not sure if this is a better world than list-non-determinism. What are we gaining from set non-det, while a lot of implementations will be list non-determinism, we are losing some valuable properties, and it is worth it? Every time i need to do a feature gate, it's a major pain, as an application developer on top of Wasm, won't use any of this tricks, maybe not worth it, gets complicated fast.
```

simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+MD: very hard to do, not the target use case for this proposal. Looking at which instructions you can test for behavior and specialize, pretty much of all them.
+
+AK: if there is a use case for qfma, support it as a principle, might as well support for other instructions?
+
+CW: main benefit for set non-det, is keeping the spec editorially not perturbed by this proposal. Trying to work this out if it is possible.
+
+DG: also minor benefit to deterministic fma. There are cases in engine we assume list determinsm, that's something the spec allows, set-non-det gives engine more leeway. From engine perspective, if a less inform application dev uses this without any sort of checking, the only thing is checking is that their algo is slow, better than not working.
```
ngzhian, line 68:
```suggestion
DG: also minor benefit to deterministic fma. There are cases in engine we assume list determinism, that's something the spec allows, set-non-det gives engine more leeway. From engine perspective, if a less informed application dev uses this without any sort of checking, the only thing is checking is that their algo is slow, better than not working.
```

simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+AK: there is a special question if we should have a separate FMA instruction.. Haven't seen the lowering?
```
ngzhian, line 70:
```suggestion
AK: there is a special question if we should have a separate FMA instruction. Haven't seen the lowering?
```

simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+DG: will call out to c intrinsics
+
+AK: there is a lowering to intrinsics, ignores all lowering, can all to c element per element, will be slow. If we add this it should not be called FMA, call it precise FMA, has to be specific qualifier to indicate so.
```
ngzhian, line 74:
```suggestion
AK: there is a lowering to intrinsics, ignores all edge case inputs, can all to c element per element, will be slow. If we add this it should not be called FMA, call it precise FMA, has to be specific qualifier to indicate so.
```

simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+CW: the kind of hazards we are imagining are people hand writing, assumes a level of deliberateness.
+
+AK: it is still a hazard, easy to miss. Like this sort example, if you do the same computation on the same data, have to be aware that it could produce different result. Increases the barrier. If QFMA produces set-non-det, we will likely see experts using intrinsics, end up writing programs per spec, not working properly, don't expect qfma returns different results. Not everyone who learns Wasm/Relaxed SIMD will write code that QFMA can result different results every time.
+
+DG: should we take a poll?
+
+ZN: maybe we should postpone Phase 4 vote
+
+DG: we can still take a vote and present it to the group when we discuss it
```
ngzhian, line 222:
```suggestion
DG: we can still take a vote and present it to the group when we discuss it

```

simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+MD: set-non-det for NaNs, not user visible, no function in C to get the payload
+
+AK: sign also
+
+MD: NaN propagation, no standard function in C to get this non-det components, not visible for programmers, what we have here is user-visible effects
+
+CW: not user-visible from a looser specification that fast math lets you do, but practically visible
+
+AK: one reason to prefer list, if we can specific FMA with single round, performance cliff. The idea of list determinism, what is the closest we can get to full determinsm without performance cliff. MVP SIMD was not doing this. Here, what is the strictest possible behavior that is feasible and useful. Impact of set-non-det in spec to user is minimal, impact in this proposal is more obvious, may be reasons to try to see how to make list works. Or something stricter?
```
ngzhian, line 174:
```suggestion
AK: one reason to prefer list, if we can specific FMA with single round, performance cliff. The idea of list determinism, what is the closest we can get to full determinism without performance cliff. MVP SIMD was not doing this. Here, what is the strictest possible behavior that is feasible and useful. Impact of set-non-det in spec to user is minimal, impact in this proposal is more obvious, may be reasons to try to see how to make list works. Or something stricter?
```

simd/2022/SIMD-10-14.md
```diff
@@ -0,0 +1,229 @@
...
+
+CW: even stricter is correlation between instructions.
+
+PP: in practice you get one result from a particular platform, you can't emulate them efficiently
+
+AK: edge cases where some specific cpus with this slow instruction
+
+PP: understand the argument that with list-determinsm, can see what results you will get, danger is that they write and test on one platform and not others
```
ngzhian, line 182:
```suggestion
PP: understand the argument that with list-determinism, can see what results you will get, danger is that they write and test on one platform and not others
```

penzn:
Could not use 'add to batch' button, though we should probably squash typo edits together with the main commit anyway.

ngzhian (approved):

ngzhian:
> Could not use 'add to batch' button, though we should probably squash typo edits together with the main commit anyway.

Yea let's squash them.

PR #1153, opened 2022-10-26
Add short talk on recovering from traps

RossTate:
I suggested putting it in the stack-switching session because it might be relevant to the discussion.


dtig (changes requested):
We're trying to guarantee as much open time for discussion as time permits for the stack switching session, as a late addition to the agenda, my recommendation is to move it to the end, and we'll try our best to get to it (either as a part of the stack switching session dynamically, or later in the day).

RossTate:
Got it. Pushed the time to later in the day.

dtig (approved):
Thanks!

PR #1193, opened 2023-01-10
Update GC-01-10.md

tlively:


zapashcanon:
@tlively here are the slides: [main.pdf](https://github.com/WebAssembly/meetings/files/10392398/main.pdf) :)

chambart:
Just to answer some questions there was about the effects of casts, I ran some tests with and without `--experimental-wasm-assume-ref-cast-succeeds` and I see a speed difference of about 10%

rossberg:
Hi @chambart, @zapashcanon, thanks so much for presenting this! Very excited about it. I'm sorry that I couldn't attend last week.

I looked at your slides, and feel reassured to see a lot of similarities to my prototype ML compiler. (I don't know if you are aware, I implemented Waml, a prototype ML compiler targeting the GC proposal, see [here](https://github.com/WebAssembly/gc/tree/iso.waml/proposals/gc/waml); an overview presentation is [here](https://github.com/WebAssembly/meetings/blob/main/gc/2021/presentations/2021-11-16-rossberg-wob%2Bwaml.pdf). Be aware, though, that I haven't yet updated it to the latest changes of the proposal yet, i.e., the cast refactoring.)

The main difference I can tell is in your use of subtyping. If I understand the discussion notes correctly, you need that deep hierarchy to model module subtyping. But why is that? As far as I know, OCaml implements module subtyping coercively, i.e., it always constructs minimal blocks when subsumption is used (which is kind of important to avoid keeping dead definitions alive, and you need it anyway due to reordering of fields). So I would expect that you'd never need Wasm-level subtyping for compiling structures. At least that's the case in my compiler: structures are compiled to Wasm structs with typed fields, but those structs do not have supertypes.

You can find the implementation of Waml type lowering [here](https://github.com/WebAssembly/gc/blob/iso.waml/proposals/gc/waml/src/lower.ml). As you can see, I only needed subtyping in two places: for closures, and for datatype constructors. Accordingly, those are also the only places where I need casts, aside from casting down from the uniform representation (eqref) when consuming a value. There is no subtype hierarchy deeper than 3.

I'd be very curious to learn why that would not work in your case.

zapashcanon:
Hi @rossberg !

I'm not the one who wrote this part but my understanding is that it's mainly because in `Flambda` (the IR we're starting from) some information is missing. In `M.n` the type of `M` is not available anymore and to be able to type access to `n` we say that `M` is a module with `n` fields or a supertype of it.

We could either:

- start from an IR where this information is still available (e.g. typedtree) but we would loose many optimizations and redo a lot of work such as pattern matching compilation;
- keep Flambda as a starting point and propagate more type information in the compiler, but for some reasons it breaks other optimizations (it was done in Flambda2 and had to be reverted).

For now we switched to arrays and have to pay bounds-checking but I believe it's not too bad.

I knew about Waml but didn't have the time to look at it. It's indeed reassuring that we ended-up making similar choices.

chambart:
Well I specifically avoided looking at Waml at first to try to avoid being influenced by the design and see if it ended up with a different solution.

Otherwise I see the need for a long subtyping chain for block in OCaml because the compiler drops most of the type informations quite early in the compiler. We could (and probably will) propagate some information into further passes but that might not be sufficient. There are untyped parts in the language that user tend to use: the 'Obj' module.
So when we access the field `n` of a block we can't really know more than that the block has 'n+1' fields. And the subtyping relation is quite good to represent that if the cast it is not too expensive.

rossberg:
@zapashcanon, @chambart, thanks for the explanation. Yeah, I can see that the lack of type information in the IR causes a problem there. That's a pity! I hope you get a chance to annotate it with more types, though I can see how that would be quite a substantial undertaking (I assume you'd need to turn all parts of Lambda and Flambda that are in the image of module lowering into a typed language).

OTOH, I think it's fine if Obj's reflection ops are more expensive.

rossberg:
@chambart, @zapashcanon, btw, are there many places where a single downcast on a block would feed into multiple projections? If not, I'd make any bet that you are better off simply using arrays to represent blocks, since a bounds check ought to be significantly cheaper than a cast (not to mention the space overhead of defining all the subtypes).

PR #1203, opened 2023-01-20
Add notes for SIMD 2023-01-20

ngzhian:


ngzhian:
@dtig are you able to get attendees list from the meeting? I forgot to keep track before I left the meeting.

PR #1258, opened 2023-04-14
Add TypeScript on WasmGC topic in GC-04-18

xujuntwt95329:


tlively:
Hi @xujuntwt95329, thanks for the PR! Since the meeting is only 1 hour long, we won't have time for both Andy's 30 minute presentation and your 40 minute presentation. Would you be free to present at the next meeting on May 2?

xujuntwt95329:
@tlively Oh sorry, I thought it's 2 hours because I saw `April 18, 9am-11am Pacific Time`, didn't notice there is `April 18, 4pm-5pm UTC`

Since May 2 is our holiday, I need to confirm if I‘m free on that time, I'll close this PR and open a new one once I confirmed the time is suitable.

Thanks for your reply

tlively:
Sorry for the typo causing confusion, @xujuntwt95329. If May 2 doesn't work, the next meeting on May 16 would also be good.

PR #1263, opened 2023-04-19
Update CG-04-25.md

conrad-watt:


conrad-watt:
I will re-target this for the following meeting and in the meantime post an interim announcement about the in-person CG.

PR #1279, opened 2023-05-17
Update GC-05-16.md

tlively:


gc/2023/GC-05-16.md
```diff
@@ -33,3 +33,176 @@ Installation is required, see the calendar invite.
...
+
+AR: In what sense are you lowering to nominal types? Because the Wasm types aren’t nominal either.
+
+JX: In the compiler we check to make sure two different types with the same structure cannot be assigned to each other.
+
+AR: But what about dynamic typing? I guess there’s no need to go too deep into it here, but right now you’re talking only about the static type system?
+
+JX: Right, the static type system is nominal.
+
+AR: Question about Any. It’s not just a top type, it’s supposed to be compatible in both directions. How can this work when you have two different representations of static and dynamic objects?
```
rossberg, line 74:
```suggestion
AR: Question about Any. In TypeScript it’s not a top type, it’s supposed to be compatible in both directions. How can this work when you have two different representations of static and dynamic objects?
```

gc/2023/GC-05-16.md
```diff
@@ -33,3 +33,176 @@ Installation is required, see the calendar invite.
...
+
+JX: Currently we want to bring more application development languages to Wasm. We don’t expect this approach will be much faster than JS.
+
+CW: Who do you imagine will use this as this first choice?
+
+JX: Developers for LT(???). In the browser, WebAssembly can be some part of the project, if not the whole project. This is more flexible than the current solution.
+
+AR: You mentioned regular typing here but the approach you are using is something close to coercive subtyping and that’s difficult. You definitely need sound subtyping to make it work. But for gradual typing you need to be more flexible, and I’m not sure how that would work with two competing implementations. With graduate typing I should be able to pass an array of C as an array of Any and vice versa. And that could be a mutable array, so you can’t just convert an array to fit the right implementation that you need. You have to keep it in this original presentation, so I’m not sure how you would support that in this approach.
```
rossberg, line 110:
```suggestion
AR: You mentioned gradual typing here but the approach you are using is something akin to coercive subtyping and that’s difficult. You definitely need sound subtyping to make it work. But for gradual typing you need to be even more flexible, and I’m not sure how that would work with two competing representations. With gradual typing I should be able to pass an array of C as an array of Any and vice versa. And that could be a mutable array, so you can’t just convert an array to fit the right implementation that you need. You have to keep it in this original presentation, so I’m not sure how you would support that in this approach.
```

gc/2023/GC-05-16.md
```diff
@@ -33,3 +33,176 @@ Installation is required, see the calendar invite.
...
+
+JX: I mentioned the gradual type here, but just want to explain the language. For the static languages, they are trying to support dynamic language features and for dynamic languages, javascript/typescript we are trying to bring more type information to their code. So for gradual typing, wasm might be a good package to bring benefits to these kinds of languages. That does not mean we are going to support all the gradual typing semantics in the languages.
+
+AR: Understood, but how do you even support the most gradual semantics with this approach. In your language, can you assign an array of C to an array of Any?
```
rossberg, line 114:
```suggestion
AR: Understood, but how do you even support the most basic gradual semantics with this approach? In your language, can you assign an array of C to an array of Any?
```

gc/2023/GC-05-16.md
```diff
@@ -33,3 +33,176 @@ Installation is required, see the calendar invite.
...
+
+JX: Not sure what the C to array means.
+
+AR: So C is some static class and you have an array of C, right? And Any basic gradual type should be able to assign that array to a variable of array of type Any. In that case, you cannot wrap the individual objects, it would be expe[snive and violate the identity of the objects because this is a mutable array so you can observe the members changing. How do you support that?
```
rossberg, line 118:
```suggestion
AR: So C is some static class and you can have an array of C, right? And with basic gradual typing you should be able to assign that array to a variable of array of type Any. In that case, you cannot wrap the individual objects, it would be expensive and in fact, violate the identity of the objects because this is a mutable array so you can observe the members changing. How do you support that?
```

gc/2023/GC-05-16.md
```diff
@@ -33,3 +33,176 @@ Installation is required, see the calendar invite.
...
+
+JX: Yes, we do rely on the engine or host environment to do externalization.
+
+CW: I imagine for a lot of users of TS, the alternative is to compile to JS, which will be pretty fast. Is the intention to compete with compiling to JS, or just to use TS in an environment where JS isn’t available?
+
+JX: Currently we want to bring more application development languages to Wasm. We don’t expect this approach will be much faster than JS.
+
+CW: Who do you imagine will use this as this first choice?
+
+JX: Developers for LT(???). In the browser, WebAssembly can be some part of the project, if not the whole project. This is more flexible than the current solution.
```
xujuntwt95329, line 108:
```suggestion
JX: Developers for IoT devices, an app-developer friendly language based on WebAssembly may help build a larger ecosystem for IoT world. In the browser, WebAssembly can be some part of the project, the part more related to UI can be still compiled to JavaScript, while other parts more related to computation can be compiled to WebAssembly, and the developer don't need to write two languages for these two parts. This is more flexible than the current solution.
```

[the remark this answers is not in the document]
xujuntwt95329: Thanks @penzn , yes the 'LT' was `IoT (Internet of Things)`, I have submitted a comment

tlively: Thanks!

gc/2023/GC-05-16.md
```diff
@@ -33,3 +33,176 @@ Installation is required, see the calendar invite.
...
+
+JX: Need performance data for the engines to support this implementation?
+
+RH: That would be one thing
+
+CW: Presumably if you are doing the faster for the user code to generate the separate getters and setters, it would probably be faster, though a little annoying. 
+
+RH: I would guess it would be in that case. but I’m sure we could put the type information in that shape so it would be really fast. Right now it’s not, so we’d need a reason to do the work like that. There might be extra metadata costs too.
+
+AR: This is a form of reflection right? It’s about putting reflection into core wasm. As you might guess, I’m not a big fan of that idea, it seems to be something that should be handled at a higher level than the basic asm instruction set.
```
xujuntwt95329, line 186: @rossberg Sorry for missing this question in the meeting, a quick reply here: 
- I think the proposed opcode `struct.get/set_indirect` is not related to reflection, it's all about getting field index from operand rather than immediate
- The idea of `interface description table` may be something like reflection, but that's not what we want to bring to the core wasm spec, it's just a feasible usage of the proposed opcode

I'll prepare an issue in `WebAssembly/gc` repo and list some benefits as well as impacts of this opcode for further discussion.

BTW, thanks for your professional questions and suggestions in the meeting, which let me realize there are more things to be considered for the type system

xujuntwt95329: @rossberg Sorry for the late reply because I didn't notice the update on this thread. Since this PR has been merged, I've created an issue https://github.com/WebAssembly/gc/issues/397 for further discussion

PR #1285, opened 2023-05-25
Update CG-06-06.md

eqrion:
@ajklein Would you be okay if presented for the follow on discussion for strings? I have a concrete proposal for compile-time imports and a string builtin API which could make the discussion more productive.

### DIFF
--- a/main/2023/CG-06-06.md
+++ b/main/2023/CG-06-06.md
@@ -25,7 +25,8 @@ Installation is required, see the calendar invite.
 1. Proposals and discussions
     1. Announcement: Next hybrid CG meeting, tentative dates in Oct 2023 [Deepti Gandluri, 5 mins]
     2. Announcement: Github roles and permissions cleanup (see #1215 and #1233) [Derek Schuff, 5 mins]
-    3. GC & strings: discussion [Adam Klein, 30 mins]
+    3. Compile time imports and string builtin API [Ryan Hunt, 25 mins]
+    4. GC & strings: discussion [25 mins]
 1. Closure
 
 ## Agenda items for future meetings


ajklein:
This works for me if we can make sure to allow sufficient time for discussion. Will the presentation fit in, say, 15min?

conrad-watt:
If this proposed agenda item is intended to communicate Mozilla's perspective on stringref and a proposed alternative design, it doesn't seem ideal to prioritise an open-ended discussion on stringref over it.

@eqrion please chime in regarding whether I've got the above correct - If I'm understanding the purpose of your presentation correctly, and if the natural length of the presentation would prevent us from having a long enough discussion, I'd prefer we have a discussion of _both_ presentations in the next again meeting (with brief recaps if needed) or more immediately through github.

eqrion:
Yeah, I'm working on a concrete proposal for an alternative here. I should have something to post on the design repo next week which should give some time for it to be discussed before the meeting.

I have a hard time estimating time, because questions during a presentation can significantly influence the time taken. A presentation without questions would probably be 25 minutes? I have a feeling there's enough discussion here to fill up the rest of the time then too.

conrad-watt:
(prompted by @eqrion's new [post](https://github.com/WebAssembly/design/issues/1479))

@ajklein would you be ok with leaving time for @eqrion to give a more full presentation in this meeting? If so, I'll reserve space in the meeting on  6-20 to ensure we have adequate follow-up discussion time.

ajklein:
I may not be able to make the 6-20 meeting, so I'm hopeful that we can fit in both @eqrion's presentation and some discussion about the higher-level question of strings & Wasm. But I agree that given the work @eqrion's put into putting together this concrete proposal, it makes sense to include it as part of the context for that discussion.

conrad-watt:
Given @eqrion's estimate above, I think we could shoot for the presentation to be timeboxed to 25 minutes, leaving 25 minutes for discussion. @eqrion this may involve making a few small cuts to your presentation in order to leave space for minor interruptions, but we can try to defer open-ended questions to the discussion period.

It seems inevitable that we're going to end this meeting with a lot of discussion left to have. The timing is somewhat awkward, because if 6-20 doesn't work for follow-ups, the next CG meeting would be 07-04, which might well be cancelled due to the US holiday.

If the above is acceptable, I'd suggest the following agenda items
```
      N. Compile time imports and string builtin API [Ryan Hunt, 25 mins]
    N+1. GC & strings: discussion [25 mins]
```
We can work out after the meeting how to approach follow-up discussions.

eqrion:
That sounds good to me. I updated the commit.

conrad-watt (approved):

PR #1287, opened 2023-05-31
Add a WASI agenda item about prototyping for resources

sunfishcode:


pchickey (approved):
Thanks! Glad we have an agenda :)

PR #1295, opened 2023-06-12
Edit time for flexible vectors, add agenda item for profiles

dtig:
Adding @penzn, @rossberg for notifications.


conrad-watt:
Is @rossberg presenting something for the profiles discussion?

dtig:
If @rossberg would like to, that would be great. The previous time we talked about profiles was as a part of the provisional Phase 4 vote for relaxed SIMD, where we decided that we need some follow on discussion on the profiles proposal, specifically around having more visibility around what the proposal introduces (IIRC we briefly talked about syntax), and have some consensus on how we decide to introduce new profiles, what profiles we see emerging in the future etc.

rossberg:
I don't have anything new to report, but I can recap.

PR #1311, opened 2023-07-03
July 17th stacks meeting

slindley:


fgmccabe:
I would have preferred a 1-1 discussion about this first. As I have said, I do not want to restart the meetings until we have several agendae set up ahead of time.

slindley:
As you know, we've been trying to arrange this one for months, but a string of holidays and other conflicts have got in the way. Moreover, we had tentatively agreed to schedule it on July 17th. I created a PR here so that people could have some warning. I knew that you had a period of leave, so might not be able to do so yourself, and I (perhaps incorrectly) assumed that that was why you hadn't responded to the latest email Daniel sent you about this.

I quite agree that it would be good to schedule the agendas for some more meetings - let's do that!

But it would seem rather perverse to defer this one even more, just because we don't have anything else lined up yet.

fgmccabe:
If you want we can have a call on this Friday.
However, scheduling a singleton meeting will defeat any benefit of scheduling Daniel's talk.

slindley:
That seems like a pretty odd claim to me... but let's take this private discussion elsewhere - it doesn't belong here. I created the PR because I thought it was entirely uncontroversial.

slindley:
@fgmccabe and I have discussed this offline. This meeting will go ahead, and we have some more meetings in the works too. Hopefully we'll have another on the 31st of July before resuming with a fuller programme from the 11th of August.

PR #1312, opened 2023-07-10
Document that phase advancement is determined by CG vote.

sunfishcode:
I noticed that CG voting was only mentioned as an entry requirement to phase 1. In practice, the CG has always voted for all phase transitions beyond phase 0.

This PR documents this existing practice.

It adds a paragraph documenting that votes are used to advance from one phase to another, and it moves the phase 1 entry requirements out of phase 0's description and into a phase 1 entry requirements section, making it more consistent with the other phases.

### DIFF
--- a/process/phases.md
+++ b/process/phases.md
@@ -21,6 +21,12 @@ In general, the process moves forward through a series of numbered phases.
 However, if issues are uncovered or consensus devolves,
 proposals should back up to the appropriate prior step.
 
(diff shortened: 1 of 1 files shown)


tlively (approved):

dschuff (approved):

Ms2ger (approved):

dtig (approved):
This now has several reviews, so I'll merge it in. Thanks for improving the document!

PR #1321, opened 2023-07-14
Update CG-07-18.md

eqrion:
I would like have a vote for phase 1 for compile-time imports and a string builtin API. Likely as a combined proposal, but that could be a topic to discuss. I think it would make sense to have this after the string discussion, as it could be affected by that.


eqrion:
Light ping for review @dtig @dschuff

dtig (approved):
Thanks for the ping.

PR #1327, opened 2023-07-21
Add agenda item for WASI-07-27 meeting

loganek:


pchickey:
Thanks, I think this is an important topic to address - I've asked @abrown, who is leading that work, to be present for that meeting.

PR #1357, opened 2023-09-07
Update WASI-09-07.md

ricochet:


pchickey:
I put these notes in the 08-24 document in #1358

PR #1377, opened 2023-09-26
Add profiles and RSIMD to CG-10.md

rossberg:


dtig:
Sorry I missed this somehow, and added a duplicate item among other agenda items for profiles. I've updated the time slot in the doc to 60 mins.

PR #1401, opened 2023-10-17
Add agenda item to WASI-10-19 for changing WASI acronym expansion

lukewagner:
The discussion, followed by a poll, is for the idea to change WASI from standing for "WebAssembly System Interface" to instead stand for "WebAssembly Standard Interfaces".  This idea has already been presented and discussed positively around the broader community, so this is an agenda item to consider making the change more formally.


pchickey:
Oops, I clicked too eagerly. We should wait for a subsequent meeting to poll on this, to give people time to digest and discuss.

PR #1414, opened 2023-10-30
Propose wasi-backtrace

abrown:


ricochet:
We may have time at the end but I think it will be close

PR #1425, opened 2023-11-09
Add poll for wasi-backtrace

abrown:
In the last meeting, I heard some support for this idea so let's put it up for a vote. A couple of people (Luke, Syrus) suggested that wasi-backtrace could or should be a core level proposal; I agree that the core spec _could_ be a good place, but WASI's recent change to become a collection of "standard interfaces" instead of "system interfaces" suggests WASI as a suitable place as well. Additionally, moving this to phase 1 makes it possible to prototype with this API under a flag and sort out the details before we need to decide exactly where this needs to live. As support improves for languages that reflect stack traces in errors, we should soon hear whether what was presented here is a good fit.

Note that @tlively floated a similar idea to this almost exactly four years ago in [WASI#159].

[WASI#159]: https://github.com/WebAssembly/WASI/issues/159


abrown:
Sorry, IDE removed some trailing spaces; I'll leave those changes in unless someone gets too confused...

abrown:
cc: @ricochet, @pchickey

fgmccabe:
I am not normally able to attend the WASI meetings. However, this item is of interest to me.
As it happens, I do not support (at the moment) this particular approach; if I understand it correctly: to import a backtrace function from the host.
My reasoning is based on the fact that (a) the form of a back trace depends on the semantics of the source language (knowledge of which is not available to the host), (b) there are some significant ambiguities about what should be in the back trace (e.g., with switching stacks, and whether JS/host frames should be included) and (c) there are other capabilities that are v. closely aligned with computing a backtrace that would not be addressed by this (e.g., displaying values of variables that are in local variables & arguments).
I feel that this should be at least surfaced to the CG as a whole as it points to a very strong need: of supporting the debugging of WebAssembly applications.

pchickey:
Based on the feedback from @lukewagner and now additionally from @fgmccabe, and our experience with wasi-threads, I think it is worthwhile to explore this topic with a larger group of stakeholders at the WASM CG level before we start a WASI proposal process. @abrown, would you be willing to bring this topic to the WASM CG and potentially to the stack switching group as well to gather more context before continuing with this as a WASI topic?

abrown:
I'm back! Here's a response for @fgmccabe: I don't follow your argument that WASI should not specify a function returning a backtrace. Regarding (a), this is a WebAssembly-level backtrace, so the host _does_ have the knowledge it needs. Regarding (b), I can't comment on the stack-switching part but I would guess that, if host frames were desired, in some later phase we could design a placeholder frame to avoid passing on host debug information. Regarding (c), I explicitly left out frame values from the initial draft but the host has the information to provide them. Note here that when this proposal was presented, it was in the context of a "host" that was a standalone WebAssembly engine, not a browser. But I'm not sure even the browser environment, given reasons a-c, precludes using a function to return a backtrace.

I would claim that wasi-threads shows the benefit of this "let's try a function" approach: provide a function with the desired behavior and, if it is successful, discuss making it a part of the core spec. IMO, wasi-threads was quite successful in (a) demonstrating that a "spawn an instance per thread" function was valuable to various sets of WebAssembly users and (b) validating that the very minimal design was good enough. And that motivated the follow-on spec work Conrad and I presented in the post-MVP threads proposal. We wouldn't have moved threads forward, I think, without the experimental feedback from wasi-threads. 

I don't really have the energy to try to pull together some consensus about the design details among a bunch of different subgroups plus the CG. That isn't really the requirement for phase 1 anyways. Perhaps we can discuss more here?

fgmccabe:
I would like to follow up a little:

- I can see a legitimate use case for an API that reports (whatever) the stack trace in the event of a crash. In fact, there are many potential use cases for being able to construct stack traces; and that is part of the problem.
- There are genuine points of ambiguity as to what should be in the trace. Depending on why you are constructing the trace (e.g., non crash reporting use cases), there will be toolchain specific variations. An example is security: some languages (looking at you Java), use the stack trace as a way of implementing security scope. (I offer no opinion as to whether that's a good idea™)
- If you mix in stack switching, you get even more variations: A JS stack trace will not reflect the chain of async function calls; but devtools does try to reflect the chain. A language like Java will link coroutines to virtual threads; which also defines the natural scope of a stack trace -- which would be different to what JS or other languages would likely do. Languages like Swift/Erlang will want to link stack traces to actors. 

My number one concern is that this space has too much variability in it to merit a simple WASI API definition; especially when you consider all the non-crash reporting use cases that folk will want. In addition, specifying a 'bad' API is something we will have to live with for the duration of wasm itself (hopefully >30years).

I do appreciate that going through the CG looks daunting and time consuming. It is daunting and time consuming.

My suggestion would be to explicitly not standardize on the API at this time (a negative intention). Of course, engine vendors are free to create their own APIs.

At the same time, I believe that there is sufficient interest in the broader community to see what can be done with stack traces and the related but different requirements for supporting stack inspection. Starting an effort in the CG seems appropriate. I believe that there may be ways of doing this without compromising security but enabling a variety of use cases (including generating stack traces for languages that want them in their exceptions).

rossberg:
Not sure if it's too late already, but I'd like to chime in as well. I agree with most of what Francis said. Making the ability to construct backtraces an unconditional and mandatory requirement is premature. Wasm deliberately does not do so so far. Forcing it through the WASI backdoor would be bad politics, since it deeply affects core Wasm.

Case in point: OCaml has recently experimented with implementing exceptions using the Wasm exception proposal. They report that this turns out to be, IIRC, two orders of magnitude (!!) slower (on V8) than instead returning a flag and performing a test and branch at every call site. The requirement to be able to report stack traces back to JS is one suspected culprit for this cost, though more investigation may be needed. Exceptions are frequently used for control flow in languages like OCaml, so their performance matters.

Of course, this is just one data point. It may well be that this could be implemented much more efficiently in an engine, with close to zero overhead. But that hasn't been demonstrated yet. Until it has been sufficiently investigated I think it is too early to standardise such a feature.

abrown:
Let's close the issue; I was only proposing this because when I needed this functionality, I really needed it. I don't have a burning desire for standardization; it just seemed like a good thing to try. @pchickey reminded me that the stack switching group may be the right group to push this kind of thing forward and that seemed like the best reason to me to stop this. (@fgmccabe, apologies for not following that subgroup!).

I can't resist making a point about the way to move this forward, though: @rossberg, I'm not sure where you get that making this a WASI API would make it "unconditional and mandatory" (it would be part of separate WIT world the user and runtime both opt in to). And I certainly did not intend it that way, just like I did not intend this to be an effort to "force it through the WASI backdoor." What I actually intended is to propose something that can be prototyped: I think a function import is the best way to bootstrap this effort. A function signature is easier to change (and in fewer places) than a new set of instructions; where else to specify that currently but in WASI? And I quite agree with @fgmccabe that "specifying a 'bad' API is something we will have to live with." But I see iterating on a function import as a good way to avoid that, by inviting feedback like "we tried to use it but, oh no!, that doesn't work for us." Perhaps consider the "function import" (WASI?) approach as you propose a different way to access stack traces?

@fgmccabe, if you do end up turning your "native intention" here into a positive one, I would be glad to support that proposal. I could use it! A simple interface may be good place to start, IMHO. Since I can't dedicate a lot of time towards it, though, let's close this PR and wait for you or someone else to move this forward.

rossberg:
> @rossberg, I'm not sure where you get that making this a WASI API would make it "unconditional and mandatory"

As long as it's not "standardised", it hopefully wouldn't. The fact that it is an WIT-level opt-in interface may not necessarily help, though: it is unlikely that this interface can be supported, even optionally, without introducing overhead in an engine implementation itself. And once such a "standard" interface exists, there will likely be significant pressure to support it in major Wasm engines.

If the current intent is just experimenting with an API then that's of course fine – that wasn't clear from the OP's framing this as a proposal. Like @fgmccabe suggests, it shouldn't move beyond experimental without consulting the core Wasm group.

Anyway, from your reply I take it that we're more or less on the same page.

SingleAccretion:
Is there a good place now, with [#159](https://github.com/WebAssembly/WASI/issues/159) closed, to register the _demand_ (and perhaps collect requirements) for this (or equivalent) feature?

fgmccabe:
A good idea™.
I suggest opening a requirements issue [here](https://github.com/WebAssembly/proposals/issues).

PR #1440, opened 2023-11-22
Add flexible vectors to CG-12-19

penzn:


dtig (approved):

penzn:
Actually maybe we can do it one meeting earlier, there seems to be enough time left there.

dtig:
Sure, please add this agenda item to that meeting instead.

dtig:
@penzn Are you still planning on presenting at the next meeting, or should this stay on 12/19?

PR #1443, opened 2023-11-30
add notes for WASI-11-30

pchickey:
@ricochet was the notetaker this meeting


ricochet (approved):
We should leave this open for a bit in case there are any corrections or additions.

PR #1454, opened 2023-12-16
2024 Jan 25: schedule wasi-observe phase 1 proposal

chrisdickinson:
Per https://github.com/WebAssembly/WASI/issues/573, requesting an agenda item to discuss `wasi-observe` on Jan 25. (I'd schedule it for Jan 11, but I saw that the preview 2 vote is on the 11th and wanted to give that more time. If you'd prefer I move this item up, I'd be happy to do so, though!)

Thanks for your consideration!


pchickey (approved):
Just got back from winter break - thank you!

PR #1460, opened 2024-01-09
WASI: Reschedule vote for WASI Preview 2 launch from 1/11 to 1/25

pchickey:
On 1/11, @gbedford will give an update on the JCO implementation progress, which is close to completion.

On 1/25, we will hear about the conclusion of the JCO implementation work, and hold the vote to launch Preview 2.

We may have to bump the wasi-observe proposal presentation and phase 1 vote to either the following meeting (WASI-02-08) or, alternatively, forward to WASI-01-11. @chrisdickinson, please let me know your preference.


ricochet (approved):
We can have another PR to move wasi-observe forward if that's the preference.

PR #1550, opened 2024-04-23
Propose agenda items for June CG meeting

tlively:
cc @aheejin for EH, @sbc100 for memory64, @rossberg for custom annotations and SpecTec, @eqrion for JS string builtins, @dtig for memory control, @SPY for FP16, @ecmziegler for compilation hints, and @tomayac for WebAssembly.org


yuri91:
If custom annotations schedules a vote for phase 4, I can do the same for branch hinting.

PR #1557, opened 2024-04-24
Move discussion of @name annotations to May 21 meeting

tlively:
Since @rossberg won't be able to attend the May 7 meeting.


tlively:
@fitzgen, would you be ok with taking 30 minutes instead of 45 minutes for the custom page size discussion so we can spend the other 30 minutes discussing name annotations?

fitzgen:
That should probably be fine. I just did 45 minutes because someone suggested that in the CG meeting two days ago.

PR #1578, opened 2024-05-20
Update CG-06.md

camio:
Add discussion on dealing with large WASM binaries

### DIFF
--- a/main/2024/CG-06.md
+++ b/main/2024/CG-06.md
@@ -108,6 +108,9 @@ Discussion of goals for website refresh
 #### WASI / Component Model update
 Status update and next steps
 
+#### Large WASM Binaries From Desktop Application Migrations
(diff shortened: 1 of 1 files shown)


tlively:
Thanks for the suggestion! A few questions for planning purposes: What kind of discussion are you planning for? Do you plan to present? How much time do you have in mind?

camio:
I could deliver a short presentation outlining the general problem, the solutions we have, and the gaps. My hope is that it would turn into a bit of a brainstorming session. I was thinking 30m-1h.

tlively:
Thanks, that sounds good. We plan to set a concrete agenda on Tuesday, so hopefully we can fit this in. Can you fix the conflict? Then I'll merge this.

camio:
@tlively Done!

PR #1590, opened 2024-05-29
Initial concrete agenda for June CG meeting

tlively:


tlively:
Also let us know if you expect to need significantly more or less time than we scheduled or if other constraints come up. We can always move things around.

PR #1593, opened 2024-05-29
Add CG-06 agenda item

rossberg:
Darn, I forgot to press the button on this one.


tlively:
I've done my best to account for this in the latest commit to #1590. I think everything should be good except for the shared-everything threads presentation not being in the most EU-friendly time.

PR #1603, opened 2024-06-07
Update CG-06-18.md

eqrion:
I'd like to present a proposal for https://github.com/WebAssembly/design/issues/1514.

### DIFF
--- a/main/2024/CG-06-18.md
+++ b/main/2024/CG-06-18.md
@@ -19,7 +19,7 @@ No registration is required for VC meetings. The meeting is open to CG members o
 
 ## Agenda items for future meetings
 
-*None*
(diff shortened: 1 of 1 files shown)


tlively:
It looks like this was accidentally put into the "future agenda items" part of the schedule. We already have 45 minutes of this meeting allocated, so @eqrion, we can either shorten this to 15 minutes or bump it to another meeting.

eqrion:
> It looks like this was accidentally put into the "future agenda items" part of the schedule. We already have 45 minutes of this meeting allocated, so @eqrion, we can either shorten this to 15 minutes or bump it to another meeting.

Ah, my bad. We should bump it to another meeting then.

tlively:
Done in #1609.

PR #1663, opened 2024-09-18
Schedule a discussion on module splitting

tlively:


conrad-watt:
Just for lack of an existing issue to discuss things in, is this presentation going to be about how module splitting requires a lot of type declarations to be duplicated?

Could this potentially be solved by repeatable sections, where we could declare that some new binary represents the _suffix_ of an existing module, so that when the new binary is instantiated the existing module's declarations are included as a prefix?

tlively:
Yes, that would be a great solution. I’ll post a design issue later today for discussion.

PR #1697, opened 2024-11-14
Create stack-2024-12-04.md

fgmccabe:
Agenda for 12/04 meeting.

PR #1702, opened 2024-11-19
add promise-calling item to stacks agenda

bakkot:
cc @fgmccabe

Possibly this is what the "Discuss issue #56" item is supposed to be? But in [the PR](https://github.com/WebAssembly/meetings/pull/1697#discussion_r1843106341) it was a different item.


bakkot:
Also, this file should presumably be named `stack-2024-12-02.md`, not `-04`, right?

PR #1735, opened 2025-01-25
[SIMD] Notes from 2025-01-17 meeting

ppenzin:


ppenzin:
@akirilov-arm PTAL as well

PR #1753, opened 2025-02-16
Add TLS discussion to THREADS-02-18.md

tlively:


tlively:
cc @conrad-watt, since we didn't finish discussing TLS in-person.

PR #1765, opened 2025-02-25
Update CG-02-25.md

tlively:


main/2025/CG-02-25.md
```diff
@@ -23,10 +23,172 @@ No registration is required for VC meetings. The meeting is open to CG members o
...
+
+TL: channeling andreas: a function call is not a constextpr
+
+LW: Neither is an import. But you can say for a fixed value of an import, then the result is constant. Same for the result of a function
+
+TL: The benefit of keeping constexprs constant and pure is that it allows lots of implementation strategies. I don’t know how this is done in practice. If everyone is doing it the same way maybe we could just standardize that.
+
+RH: I think you could still do lazy instantiation. My fear about doing global init is that [... ]
+
+CW: Why is it important that this goes in this way …
```
tlively, line 104: @eqrion, @conrad-watt, do you remember what you said here?

eqrion: "I think you could still do lazy initialization of globals in the case where none of the global initializers do any calls to imports. My fear about allowing global initializers to call imports is that a global initializer could get a funcref, call an import with that funcref, then have that import call back into the module and observe uninitialized globals"

dschuff (approved):

PR #1774, opened 2025-03-05
Update CG-03-25.md

guybedford:
As a follow-up to next weeks discussion on the `wasm:` scheme it would be good to discuss more generally the JS ESM integration builtin interop story and gauge current opinions on where things may end up post ESM-integration.


tlively:
Since @eqrion wasn't able to present today, should we combine the initial discussion of the `wasm:` scheme into this?

guybedford:
I think it would still be good to keep the items separate. I could possibly take 5 mins off the timebox here if 15 minutes is still needed for that discussion in this meeting - @eqrion?

guybedford:
Let's maybe land this as-is for now, and then change the timebox if necessary if we get the other agenda item reposted, otherwise that follow-up can remain separate.

PR #1775, opened 2025-03-10
Poll on the relationship between JSPI and core-stacks

kmiller68:
Order isn't particularly important, happy to go first if that makes more sense.

### DIFF
--- a/stack/2025/stack-2025-03-10.md
+++ b/stack/2025/stack-2025-03-10.md
@@ -12,7 +12,7 @@
 
 1. Ask me anything about JSPI. Discussion led by Francis McCabe
 2. Informal poll on the relationship between JSPI and core-stacks by Keith Miller (15m)
-    1. "Does JSPI's existance preclude a core-stack switching design is not side-stack based"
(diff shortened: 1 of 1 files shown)


conrad-watt:
If you're intending to make a presentation along with the poll, could you provide a timebox estimate?

Also, could you add some additional detail on what you're intending to poll? The precise wording would be ideal.

kmiller68:
Done.

stack/2025/stack-2025-03-10.md
```diff
@@ -11,6 +11,8 @@
 ## Agenda items
 
 1. Ask me anything about JSPI. Discussion led by Francis McCabe
+2. Informal poll on the relationship between JSPI and core-stacks by Keith Miller (15m)
+    1. "Does JSPI's existance preclude a core-stack switching design is not side-stack based"
```
conrad-watt, line 15:
```suggestion
    1. "Does JSPI's existence require that any future core-stack switching design is side-stack based?"
```

conrad-watt (approved):
Just a small nit to remove the double negative.

@fgmccabe feel free to merge the PR if you're comfortable with this agenda change

PR #1845, opened 2025-07-01
Update WG-07-23.md

eqrion:
Per the discussion in #1844, we should move the rest of these proposals to phase 5 while we're at it.


eqrion:
Sounds good, I added threads too.

PR #1867, opened 2025-07-30
Add agenda item for WG to advance Threads proposal to phase 5

Achierius:
From my understanding, three major browser engines (JavaScriptCore, V8, Spidermonkey) have implemented this spec, and it's been stable for around a year now. If there's no further spec work to be done, then we should move it to stage 5. Since there may already be a phase 4 vote for CSP, this seems as good a date as any. 

From reading some past discussions ([e.g.](https://github.com/WebAssembly/meetings/pull/1844)) it doesn't seem like there's a requirement for the sponsor (in this case Conrad Watt) to be the one to add this to the agenda.

### DIFF
--- a/main/2025/CG-08-12.md
+++ b/main/2025/CG-08-12.md
@@ -16,6 +16,7 @@ No registration is required for VC meetings. The meeting is open to CG members o
 1. Proposals and discussions
    1. Update on WebAssembly's part of CSP (Francis McCabe)
       (Potential vote for phase 4.)
+   2. Advance Threads proposal to Phase 5 (Marcus Plutowski)
 1. Closure
 
 ## Agenda items for future meetings


Achierius:
cc @tlively @conrad-watt

conrad-watt:
Hi @Achierius - threads not moving to phase 5 is related to the ongoing attempt to transition to a SpecTec-based spec (https://github.com/WebAssembly/meetings/pull/1844#issuecomment-3009156081). Currently we're waiting to see if the next Wasm 3.0 spec release can be based on SpecTec (@rossberg can say more about progress here). If this goes through, we'll then need to extend SpecTec with some additional features to support threads before the threads specification can be merged in.

This PR does raise an interesting point - if we're serious about transitioning to SpecTec, we should explicitly mention SpecTec requirements into our process. I think mentally everyone has been waiting to see if the 3.0 transition works before doing this, but it probably makes sense to have some discussion on this now. I've opened https://github.com/WebAssembly/meetings/issues/1868, and included some practical points about threads phase advancement at the end.

PR #1899, opened 2025-10-02
Add wasm_of_ocaml experiential talk

rickyvetter:
We are interested in talking about wasm_of_ocaml, how it works and the challenges with building it, and how it's used in production today. With @vouillon @oliviernicole


tlively (approved):
Thanks, sounds great!

Issue #1891, opened 2025-09-23
Agenda for October in-person meeting


tlively:
Hi everyone,

It's time to start collecting agenda items and time constraints for the October meeting. Feel free to make PRs directly to the [agenda doc](https://github.com/WebAssembly/meetings/blob/main/main/2025/CG-10.md) or just respond to this issue.

ccing proposal champions @rossberg @conrad-watt @fgmccabe @guybedford @alexcrichton @echamudi @fitzgen @slindley @KloudKoder @ecmziegler @lukewagner @ppenzin @dtig @bvisness @wingo @abrown @zapashcanon @chambart @eqrion @sjrd

sjrd:
* Update on JS Primitive Builtins, and phase 2 vote (30 minutes), @sjrd

I'll try to be there in person, but if I do I will have to leave fairly early on Wednesday. So preferably on Tuesday.

rossberg:
Would there be interest in having another discussion about standardising profiles, now that informal takes like lime1 start popping up?

fitzgen:
Is there interest in a general update on custom page sizes and implementation progress?

tomayac:
- I can give a quick update on [`<wasm-compat>`](https://webassembly.org/news/2025-09-17-wasm-compat/) and also on the [News](https://webassembly.org/news/) section in general. @rossberg's article was number one on Hacker News, so we may want to make use of this section more for disseminating Wasm news.
- General [website Issues](https://github.com/WebAssembly/website/issues).
- I also wanted to ask if it's OK with the community to hook the News section up with the [RSS feed](https://feeds.libsyn.com/520688/rss) of the [WasmAssembly podcast](https://www.youtube.com/playlist?list=PLNYkxOF6rcIA46I-YCX3ASF4SRb548z8s) so new episodes get cross-posted on the site.

(Scheduling constraint: I'm only there on day 1 of the meeting.)

rickyvetter:
Added a potential agenda item at #1899

tlively:
I added all the suggested agenda items to the agenda doc in https://github.com/WebAssembly/meetings/pull/1911. For any further suggestions, please create new PRs or comment on existing PRs to the agenda doc.

PR #1902, opened 2025-10-02
Add website-related topics to agenda

tomayac:
Context: #1891


tomayac:
> Oh, how long do you think we'd need for this?

30min maybe. Added to the PR.

PR #2017, opened 2026-04-12
Add "JS Text Encoding Builtins" to the agenda

daxpedda:
I would like to present the [JS Text Encoding Builtins proposal](https://github.com/WebAssembly/design/issues/1583) in the next meeting.


tlively (commented):
There are only 20 minutes free in the schedule. Is that enough time for you?

daxpedda:
> There are only 20 minutes free in the schedule. Is that enough time for you?

I honestly don't know, but I assume its not a problem either. If something was left to discuss we can follow up on that.
So lets just do with what we have!

tlively (approved):

PR #2037, opened 2026-05-07
Add notes from CG 2026-04-21

dschuff:


main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+BT: we aren’t doing FP atomics?
+
+CW: a separate discussion, I forget the rationale.
+
+TL: for next steps, I’d love to get KW’s input on the interpreter question. We can of course add 66 new instructions if we have to, but I’d prefer not to.
+
+
+#### [JS Text Encoding Builtins](https://github.com/WebAssembly/design/issues/1583) (daxpedda, 20 minutes)
+
+TODO: hackmd link.
```
dschuff: @daxpedda do you have any other links you want in the notes, or is the link to the issue sufficient?

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+
+BT: yeah those are already not too fun
+
+KM: don’t you have to look up the type anyway already?
+
+BT: I think we cheat and rewrite struct.get in place to an internal opcode
+
+TL: for v8 we're not doing in-place interpretation, fewer instructions would definitely better
+
+KM: we do do in-place interpretation. I’d have to get back to you on what the consequences would be for that. If it’s the same as whad we do for regular structs, it might not be so bad.
```
tlively:
```suggestion
KM: we do do in-place interpretation. I’d have to get back to you on what the consequences would be for that. If it’s the same as what we do for regular structs, it might not be so bad.
```

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+
+AR: yeah you don’t have to compare those.
+
+TL: on what the rules should be, I agree with CW that as long as you don’t add new type variables into the principal types rule, it should be good. For where it “doesnt matter” what the type is for parametric cases... of course it does matter for real implementations. Would be interested to hear more about how it affects interpreters.
+
+BT: we aren’t doing FP atomics?
+
+CW: a separate discussion, I forget the rationale.
+
+TL: for next steps, I’d love to get KW’s input on the interpreter question. We can of course add 66 new instructions if we have to, but I’d prefer not to.
```
tlively:
```suggestion
TL: for next steps, I’d love to get KM’s input on the interpreter question. We can of course add 66 new instructions if we have to, but I’d prefer not to.
```

main/2026/CG-2026-04-21.md
daxpedda: It is. But for reference I think it doesn't hurt to include the HackMD: https://hackmd.io/90QpK3FCR0-q1AZQh_udKA.

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+
+
+#### [JS Text Encoding Builtins](https://github.com/WebAssembly/design/issues/1583) (daxpedda, 20 minutes)
+
+TODO: hackmd link.
+
+This proposes 5 functions: one for decoding from GC arrays and one from linear memory, another for measuring in UTF-8 (in JS, the length is in WTF16, so we need a separate function). And one for encoding for GC arrays and for LM. the goal is a more performant way for toolchains to decode and encode strings from their native format to JS string (in the form of an externref). 
+There are a bunch of open questions. One particular one is how to deal with wasm64. An easy solution but maybe controversial is to make everyting an i62 insstead of 32, e.g. the length. Then a user who doesn’t want to deal with bigger things can just not do that, and then not worry about trapping. But we need the i64 for encoding and decoding, to handle addresses that might be bigger than i32.
```
tlively:
```suggestion
There are a bunch of open questions. One particular one is how to deal with wasm64. An easy solution but maybe controversial is to make everything an i64 instead of 32, e.g. the length. Then a user who doesn’t want to deal with bigger things can just not do that, and then not worry about trapping. But we need the i64 for encoding and decoding, to handle addresses that might be bigger than i32.
```

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+CW: a separate discussion, I forget the rationale.
+
+TL: for next steps, I’d love to get KW’s input on the interpreter question. We can of course add 66 new instructions if we have to, but I’d prefer not to.
+
+
+#### [JS Text Encoding Builtins](https://github.com/WebAssembly/design/issues/1583) (daxpedda, 20 minutes)
+
+TODO: hackmd link.
+
+This proposes 5 functions: one for decoding from GC arrays and one from linear memory, another for measuring in UTF-8 (in JS, the length is in WTF16, so we need a separate function). And one for encoding for GC arrays and for LM. the goal is a more performant way for toolchains to decode and encode strings from their native format to JS string (in the form of an externref).
```
tlively:
```suggestion
This proposes 5 functions: one for decoding from GC arrays and one from linear memory, another for measuring in UTF-8 (in JS, the length is in WTF16, so we need a separate function). And one for encoding for GC arrays and for linear memory. the goal is a more performant way for toolchains to decode and encode strings from their native format to JS string (in the form of an externref). 
```

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+CW: any thoughts on this? Maybe RH?
+
+RH: overall it looks reasonable. I have ideas we can discuss later. I’m conflicted on the 64 bit thing, there are merits to both. Having them all as i64 is annoying. Also having multiple results is annoying especially when calling builtins. We’ll have to support both anyway. No great ideas right now.
+
+JK (chat): in practice, browsers cap JS strings at somewhere around 2**28 to 2**30. I expect that to remain the case for a loooong time, because to bump it you'd have to make all strings bigger, and given how common strings are (and how many of them are short) that'd be hugely expensive for all existing code/websites out there.
+
+
+
+
+GB: one decision here: the problem of how to reference the memory, when you encode into LM, how to reference it. The current design uses an externref to represent the wasm memory. It would be good to get feedback on that for future phases.
```
tlively:
```suggestion
GB: one decision here: the problem of how to reference the memory, when you encode into linear memory, how to reference it. The current design uses an externref to represent the wasm memory. It would be good to get feedback on that for future phases.
```

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+
+TL: you could use export names, that would be robust to merges and splits.
+
+RH: or have a dedicated slice type that refers to memory or GC array, or a memory ref. But those are bigger changes with tradeoffs.
+
+DP: what would be slow? Getting the externref or passing it?
+
+RH: you need to make sure the ref is an object, then cast to WA.memory. Then you have to find the most current data pointer and bounds length. Since WA.memory wraps an array buffer thats a couple indirections. Maybe you could cache, i’m not sure. But the naive one would be multipel casts and indirections.
+
+DP: but I would imagein that any way to get these would have those downsides.
```
tlively:
```suggestion
DP: but I would imagine that any way to get these would have those downsides.
```

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+
+RH: yeah the casting is inherent if you use an externref. Some schemes would bind the memory directly so there’s no casting.
+
+SD: if we add core conversion from GC arrays to linear memory and back, core insts would have internal access. If we could create GC arrays from LM content, you could build the LM version of the builtins using the GC versions. You could locally recognize that the GC array doesn’t actually need to be allocated so it coudl be easier to make that fast than the externref-memory one.
```
tlively:
```suggestion
SD: if we add core conversion from GC arrays to linear memory and back, core insts would have internal access. If we could create GC arrays from linear memory content, you could build the linear memory version of the builtins using the GC versions. You could locally recognize that the GC array doesn’t actually need to be allocated so it could be easier to make that fast than the externref-memory one. 
```

main/2026/CG-2026-04-21.md
```diff
@@ -30,6 +30,231 @@ To be filled in after the meeting.
...
+
+DP: but I would imagein that any way to get these would have those downsides.
+
+RH: yeah the casting is inherent if you use an externref. Some schemes would bind the memory directly so there’s no casting.
+
+SD: if we add core conversion from GC arrays to linear memory and back, core insts would have internal access. If we could create GC arrays from LM content, you could build the LM version of the builtins using the GC versions. You could locally recognize that the GC array doesn’t actually need to be allocated so it coudl be easier to make that fast than the externref-memory one. 
+
+DP: the main concern is that we want to avoid copying. You did argue that that could be optimized, but it would be preferable not to have to rely on that optimization. From the toolchian perspective, if we assume that e.g. rust one day supports all of the GC stuff, the only way currently to do that is inline asm which is unstable but those instructions might not always be right next to each other, so that could hinder the optimization. But it seems less than ideal to rely on it.
```
tlively:
```suggestion
DP: the main concern is that we want to avoid copying. You did argue that that could be optimized, but it would be preferable not to have to rely on that optimization. From the toolchain perspective, if we assume that e.g. rust one day supports all of the GC stuff, the only way currently to do that is inline asm which is unstable but those instructions might not always be right next to each other, so that could hinder the optimization. But it seems less than ideal to rely on it.
```

tlively (approved):

PR #2114, opened 2026-08-14
Add a vendor-neutral skill for producing notes from transcripts

tlively:
Add a skill that runs a sequence of python scripts to process Zoom or Meet transcripts into meeting notes. The skill can be used by any agent and the python scripts can be configured to call out to any LLM provider. The general strategy is to do as much of the processing as possible deterministically, and do the rest by giving the LLM only a small fragment of the document at a time, preventing the LLM from summarizing away important information. The skill also instructs the agent to ask for human review after each step of the processing pipeline.

### DIFF
--- a/skills/wasm-meeting-notes/SKILL.md
+++ b/skills/wasm-meeting-notes/SKILL.md
@@ -1,0 +1,79 @@
+---
+name: wasm-meeting-notes
+description: Convert raw meeting transcripts (Google Meet or Zoom) into comprehensive meeting minutes in the WebAssembly community group style using an interactive pipeline.
+---
+
+# Wasm Meeting Notes
+
+## Overview
+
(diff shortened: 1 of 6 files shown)


bvisness:
What does this mean for the handwritten meeting notes? I'm a bit concerned that, if those are not part of the pipeline, we will get poor results because the raw transcripts are likely to completely butcher technical terms.

tlively:
Indeed, this is not a "push button, upload notes" kind of workflow and still requires a significant amount of editing work. For example doing replace-all to fix butchered jargon and replace meeting room names with people names, but also to make further cleanups and fixes that the AI didn't quite get right. But the work is still a lot easier than if you were to do it manually starting from a raw transcript, and the results are a lot more complete than just relying on handwritten notes.

tlively:
I still need to experiment with using handwritten notes to guide the AI generation of the notes from the transcript. This will be important for e.g. in-person meetings where many speakers will not be differentiated in the transcript.

PR #2125, opened 2026-09-02
Add meta-descriptors discussion to CG-2026-09-22.md

tlively:


titzer (commented):
FWIW this semester the CG meetings overlap lecture; so scheduling this later in the slot is better.

